PR #10, opened 2021-03-04
Add retries to the install subcommand polling

### LINKED ISSUE #9: Add retries to polling
When the CLI waits for the user to complete the cluster setup flow, it polls once per second. If one of those requests fails, it exits immediately:

```
Waiting for agent setup completion...
Error: setup failed, unexpected HTTP status code 502 for URL https://buoyant.cloud/connect-agent?linkerd-buoyant=Y8_FKaEaFoDrwVI8
Usage:
  linkerd-buoyant install [flags]
```

The CLI should instead retries at least once when this happens.

siggy:
The `install` subcommand polls Buoyant Cloud once per second while the
user sets up their cluster in their browser. If a single request failed,
the CLI would exit.

Modify the `install` polling to only exit after 3 consecutive request
failures.

Fixes #9

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -20,6 +20,8 @@ import (
 // a browser window.
 type openURL func(url string) error
 
+const maxPollingRetries = 3
+
 func newCmdInstall(cfg *config) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "install [flags]",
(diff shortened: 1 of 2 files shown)


cmd/install.go
```diff
@@ -139,11 +141,17 @@ func newAgentURL(cfg *config, openURL openURL) (string, error) {
...
 		if resp.StatusCode == http.StatusAccepted {
 			// still polling
+			retries = 0
 			time.Sleep(time.Second)
 			continue
 		}
 
 		if resp.StatusCode != http.StatusPermanentRedirect {
+			retries++
+			if retries < 3 {
```
zaharidichev, line 151: Can we make that a const?

zaharidichev (approved):
LGTM. Hope we do not hit that code path :)

rmars (approved):
🚀 🚀 🚀  nice test!

PR #11, opened 2021-03-05
Revert to older version of pkg/browser

### LINKED ISSUE #3: Investigate install subcommand using a text-based browser
# Background

User reported `linkerd buoyant install | kubectl apply -f -` fails due to extra text printed to stdout:
Screenshot 2021-02-22 at 20 07 55

Removing the pipe to kubectl yields a text-based browser:
Screenshot 2021-02-22 at 20 07 43

# Observations

We currently use github.com/pkg/browser to detect and open a browser window. It's possible this package is attempting to open a text-based browser.

`linkerd dashboard` works similarly, though linkerd is on a version 3.5 years old:
https://github.com/linkerd/linkerd2/blob/ed65a96f43f1e9e9410299de51ade2a205a04f3e/go.mod#L32

Confirmed github.com/pkg/browser recently added support for launching browsing other than `xdg-open`:
https://github.com/pkg/browser/pull/26

# Proposal

A few options:
1) Revert this repo to an older version of https://github.com/pkg/browser
2) Submit a PR to https://github.com/pkg/browser
3) Roll our own https://github.com/pkg/browser
4) Remove browser-opening functionality, require the user to copy a URL

I think option one is the best solution in the medium-term, as it's very little effort and does not sacrifice functionality. Longer term we may want to evaluate option 2, or, failing that, option 3.

siggy:
pkg/browser#26 added support for opening URLs in text-based browsers,
such as lynx. Unfortunately this does not play well with the CLI's
`linkerd-buoyant install | kubectl apply -f -` pattern, as we intend for
only a k8s manifest to be piped to kubectl.

Revert to an older version of `pkg/browser` that only supports
`xdg-open`.

Fixes #3


zaharidichev (approved):
I tested at on my machine (which was patient zero) and it works

rmars (approved):
🧓 🌟

PR #16, opened 2021-05-25
Add API to transfer Linkerd certs data

zaharidichev:
This change augments the agent's functionality to send Linkerd TLS cert data
to the bcloud api. This is done through a new API endpoint called `LinkerdInfo`.
We envision that this API will serve to transfer data other than certificates in the 
future. For now we can transfer three type of certs objects - control plane certs, 
proxy certs and webhook certs. This PR adds implementation to send only control
plane certificates.

To test: 
1. Bring up your docker compose env.
2. Create a kind cluster and install Linkerd on it.
3. Build the agent
`docker build -f agent/Dockerfile -t ghcr.io/buoyantio/linkerd-buoyant:certs .`
4. Load the image onto the kind cluster
 `kind load docker-image ghcr.io/buoyantio/linkerd-buoyant:certs`
5. Add a cluster
6. Get your agent manifest, but do not apply it as you need to make some edits
7. Edit the agent image to point to `linkerd-buoyant:certs` and the pull policy to `IfNotPresent`
8. Add a ports skip annotation to the bcloud agent deployment spec.
That needs to go in `deployment.spec.template.metadata.annotations`
  ```
  config.linkerd.io/skip-outbound-ports: 4191
  ```
9. Now you can apply the agent manifest
10. Tail the log of the agent: 

```
 kubectl logs -n buoyant-cloud <agent-pod> buoyant-cloud-agent  -f | grep handleLinkerdInfo
```

Note that this will not be very interesting because the default LD installation makes the roots
and the issuer the same: https://github.com/linkerd/linkerd2/blob/main/cli/cmd/options.go#L673. 
If you want to make the experience more exciting, you can supply your own certs or even configure
cert manager to rotate them periodically


go.mod
```diff
@@ -9,7 +9,6 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/cobra v1.1.3
 	google.golang.org/grpc v1.37.0
-	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0 // indirect
```
adleong, line 12: Running `bin/gen-proto` brings these lines back, but running `go mod tidy` deletes them.  I think the consensus is that they should be left in.

adleong: Actually, I think this can be fixed by adding a `tools.go`:

```console
> cat tools.go
// +build tools

package main

import (
	_ "google.golang.org/grpc/cmd/protoc-gen-go-grpc"
)
```

adleong (approved):
Looks good!

agent/main.go
```diff
@@ -150,13 +151,15 @@ func main() {
...
+	go linkerdInfoHandler.Start()
 
 	// run admin server
 	go admin.StartServer(*adminAddr)
 
 	// wait for shutdown
 	<-stop
 	log.Info("shutting down")
-	workloadHandler.Stop()
+	//workloadHandler.Stop()
```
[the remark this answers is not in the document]
zaharidichev: Ah, my bad... I was just testing some things

agent/pkg/k8s/certificates.go
```diff
@@ -0,0 +1,198 @@
...
+	controlPlaneComponentLabel   = "linkerd.io/control-plane-component"
+	identityComponentName        = "identity"
+	linkerdProxyContainerName    = "linkerd-proxy"
+	linkerdRootsEnvVarName       = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS"
+	proxyAdminPortName           = "linkerd-admin"
+	linkerdNsEnvVarName          = "_l5d_ns"
+	linkerdTrustDomainEnvVarName = "_l5d_trustdomain"
+)
+
+func (c *Client) GetControlPlaneCerts() (*pb.ControlPlaneCerts, error) {
```
zaharidichev, line 28: Btw, I think, a good way to test that end to end is to supply a pod in the fake k8s API with a container IP that is localhost with some port, and spin a fake TLS server locally that will return the expected issuer certs when connected to. I actually plan to attempt that once I get some initial feedback.

siggy (commented):
overall this is looking good. i left some questions around the `.proto` file. i plan to do another pass looking closer at the implementation, but wanted to make sure all proto changes were nailed down first. 👍

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,53 @@ message Event {
...
+  CertData crt_chain = 1;
+  string component_name = 2;
+}
+
+message CertificateInfo {
+    oneof info {
+    ControlPlaneCerts control_plane = 1;
+    ProxyCerts proxy = 2;
+    WebhookCerts webhook = 3;
+  }
```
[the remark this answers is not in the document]
zaharidichev: If you ask for the pod metadata, the only thing that can be derived is the roots.

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,53 @@ message Event {
...
+}
+
+message LinkerdMessage {
+  Auth auth = 1;
+  oneof message {
+    CertificateInfo crt_info = 2;
```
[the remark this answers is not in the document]
zaharidichev: We can send multiple pod related messages at once, and I think that is a good change

agent/pkg/k8s/certificates.go
```diff
@@ -0,0 +1,198 @@
...
+
+const (
+	controlPlaneComponentLabel   = "linkerd.io/control-plane-component"
+	identityComponentName        = "identity"
+	linkerdProxyContainerName    = "linkerd-proxy"
+	linkerdRootsEnvVarName       = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS"
+	proxyAdminPortName           = "linkerd-admin"
+	linkerdNsEnvVarName          = "_l5d_ns"
+	linkerdTrustDomainEnvVarName = "_l5d_trustdomain"
+)
```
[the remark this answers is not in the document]
zaharidichev: will try

agent/pkg/handler/linkerd_info.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/buoyantio/linkerd-buoyant/agent/pkg/api"
+	"github.com/buoyantio/linkerd-buoyant/agent/pkg/k8s"
+	pb "github.com/buoyantio/linkerd-buoyant/gen/bcloud"
+	log "github.com/sirupsen/logrus"
+	"google.golang.org/protobuf/encoding/prototext"
+)
+
+const (
+	linkerdInfoInterval = 10 * time.Second
+)
```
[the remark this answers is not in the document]
zaharidichev: We can listen for issuer updated event issues by the identity service. Nevertheless this 10 secs interval is too tight. We can get away with something like a few days even, I think

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,69 @@ message Event {
...
+  bytes raw = 1;
+}
+
+// Represents the certificates that the control plane
+// has been configured with.
+message ControlPlaneCerts {
+  // This is the identity issuer certificate chain
+  CertData issuer_crt_chain = 1;
+  // The roots that the control plane has been configured with
+  CertData roots = 2;
```
olix0r, line 112: Just to put a finer point on this: are these the roots that the control plane will add to newly created proxies? Or the root (singular) that signed the issuer? Or something else?

zaharidichev: This is the roots that the control plane will add to newly created proxies - the roots that Linkerd has been configured with (which should be on the config map).

olix0r: Yeah. Perhaps worth calling that out in more detail in the comment?

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,69 @@ message Event {
...
+// roots or a chain of certificates.
+message CertData {
+  bytes raw = 1;
+}
+
+// Represents the certificates that the control plane
+// has been configured with.
+message ControlPlaneCerts {
+  // This is the identity issuer certificate chain
+  CertData issuer_crt_chain = 1;
```
olix0r, line 110: take it or leave it: does it make sense to encode each cert in its own cerdata? i.e. so a chain is a `repeated CertData ...`? This is what we do in the proxy-api: https://github.com/linkerd/linkerd2-proxy-api/blob/a7f2efd04dc2905e5591bb68933cb234d39f09c2/proto/identity.proto#L34-L39

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,69 @@ message Event {
...
+}
+
 //
 // API
 //
 
 service Api {
   rpc WorkloadStream(stream WorkloadMessage) returns (Empty) {}
   rpc AddEvent(Event) returns (Empty) {}
+  rpc LinkerdInfo(LinkerdMessage) returns (Empty) {}
```
olix0r, line 159: I don't have a lot of context, but would it make sense for each of these message types (i.e. each type of certificate info) to have its own api endpoint, rather than vary them within the body? It probably doesn't really matter, but if we considered a router/proxy in between, it would actually have to know about the message body to make decisions. If they're split into different API endpoints, they can have separate success rates, policies, etc....

olix0r (commented):
my comments are just food for thought, generally looking good to me!

proto/buoyant-cloud-api.proto
zaharidichev: Makes sense, initially I had it like that because I wanted to avoid parsing the certificate data and juts dump the raw bytes. This has proved to be hard as I get an array of certs from the proxy, and I encode them to a pem blob, ultimately using the utils provided in the linkerd repo. So yes, I might as well have them as separate certs..

agent/pkg/k8s/certificates.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if l5dtrustdomain == "" {
+		return "", fmt.Errorf("could not find %s env var on proxy container [%s]", linkerdTrustDomainEnvVarName, container.Name)
+	}
+	return fmt.Sprintf("%s.%s.serviceaccount.identity.%s.%s", podsa, podns, l5dns, l5dtrustdomain), nil
+}
+
+func extractRootsCerts(container *v1.Container) ([]*pb.CertData, error) {
+	for _, ev := range container.Env {
+		if ev.Name == identity.EnvTrustAnchors {
```
siggy, line 121:
```go
if ev.Name != identity.EnvTrustAnchors {
  continue
}
```

agent/pkg/k8s/certificates.go
```diff
@@ -0,0 +1,178 @@
...
+	sn, err := getServerName(pod.Spec.ServiceAccountName, pod.ObjectMeta.Namespace, container)
+	if err != nil {
+		return nil, err
+	}
+
+	dialer := new(net.Dialer)
+	dialer.Timeout = 5 * time.Second
+
+	conn, err := tls.DialWithDialer(
+		dialer,
```
siggy, line 154: how about:
```go
conn, err := tls.DialWithDialer(
	&net.Dialer{Timeout: 5 * time.Second},
```
... unless it makes sense to start re-using the dialer on every iteration?

agent/pkg/k8s/certificates.go
```diff
@@ -0,0 +1,178 @@
...
+		"tcp",
+		fmt.Sprintf("%s:%d", pod.Status.PodIP, port), &tls.Config{
+			InsecureSkipVerify: true,
+			ServerName:         sn,
+		})
```
siggy, line 159: i assume this is to subvert linkerd's existing tls? mind leaving a comment describing what's happening? (`InsecureSkipVerify: true` can look scary ;) )

agent/pkg/k8s/certificates.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// skip the end cert
+	certs := conn.ConnectionState().PeerCertificates
+	if len(certs) < 2 {
+		return nil, fmt.Errorf("expected to get at least 2 peer certs, got %d", len(certs))
+	}
+
+	certificates := certs[1:]
+	certsData := make([]*pb.CertData, len(certificates))
+	for i, crt := range certs[1:] {
```
siggy, line 172: should this be...?
```go
for i, crt := range certificates {
```

siggy (commented):
this is coming together very nicely! left some minor comments...

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,69 @@ message Event {
...
+// A message that carries one or more pod certificates
+message ProxyCerts {
+  repeated ProxyCert certs = 1;
+
+  message ProxyCert {
+    // The peer chain which include the issuer
+    // cert as well as the leaf certificate
+    CertData peer_crt_chain = 1;
+    // The roots this proxy has been configured with
+    CertData roots = 2;
```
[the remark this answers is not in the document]
zaharidichev: They are not.  The ControlPlaneCerts.issuer_crt_chain does not contain the leaf cert, while the ControlPlaneCerts.peer_crt_chain, does. I guess we can use one message for both and just have an unwritten contract that the proxy certs contain the leaf as well. 🤷  What does the consolidation of messages bring us in the general case?

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,69 @@ message Event {
...
+    string pod_name = 3;
+    string namespace = 4;
+    string workload_name = 5;
```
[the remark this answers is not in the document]
zaharidichev: Doesn't the workload message pack the whole .. workload. Like all the metadata, etc etc. Maybe I am missing something but wouldn't that end up transferring quite a bit of data we do not need?.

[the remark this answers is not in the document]
zaharidichev: Also, a workload message packs all the pods, we care about a specific pod here, right?

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,69 @@ message Event {
...
+  CertData issuer_crt_chain = 1;
+  // The roots that the control plane has been configured with
+  CertData roots = 2;
+}
+
+// A message that carries one or more pod certificates
+message ProxyCerts {
+  repeated ProxyCert certs = 1;
+
+  message ProxyCert {
```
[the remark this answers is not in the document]
zaharidichev: We are not really sending a list of all pods. Just their names, namespaces, workload kinds and workload names. I _think_, this is significantly less than what is in being sent on the workload stream.  


> do we want two parallel APIs that are sending up a list of all pods on the cluster? how do we reconcile if they disagree? also we don't necessarily need to decide this right now.

I think these are separate APIs, serving separate purpose and from neatness perspective, I lean on leaving them separate and not mixing them. I however might be missing something. Is there very good reason to try and interleave them? Wrt to getting out of sync, I think even if this happens, given a bit of time things will converge. For example, say a
`ProxyCerts` message arrives before bcloud has received info from the workload stream for  workload/pod.  The certs message will just be ignored because we do not know about this workload/pod. On next cert update iteration, enough time should have passed that the Workload update/refresh has arrived and we already know about this pod, so we insert the certs data. I think in general any of these races that stem from timing issues will get resolved after a few iterations at most. Correct me if I am wrong. In essence if we see a certs update message we only write to the table, we never delete anything. So the only way to get out of sync if we receive it before we know about the pod. And that should get fixed by itself given enough time.

proto/buoyant-cloud-api.proto
```diff
@@ -92,11 +92,67 @@ message Event {
...
+    repeated CertData roots = 2;
+    string pod_name = 3;
+    string namespace = 4;
+    string workload_name = 5;
+  }
+}
+
+message WebhookCerts {
+  // The certificate chain for this webhook cert
+  CertData crt_chain = 1;
```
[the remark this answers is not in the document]
zaharidichev: It would be. @olix0r the web hook certs can contain a chain of certs, correct?

agent/pkg/k8s/certificates.go
```diff
@@ -0,0 +1,182 @@
...
+	sn, err := getServerName(pod.Spec.ServiceAccountName, pod.ObjectMeta.Namespace, container)
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := tls.DialWithDialer(
+		&net.Dialer{Timeout: 5 * time.Second},
+		"tcp",
+		fmt.Sprintf("%s:%d", pod.Status.PodIP, port), &tls.Config{
+			// we want to subvert TLS verification as we not need
```
siggy, line 155: ...we *do* not need...

siggy (commented):
looks good! i plan to give it one more pass after webhook and pod proto has been removed.

proto/buoyant-cloud-api.proto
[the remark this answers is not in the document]
olix0r: It's possible. probably most likely via cert-manager.

agent/main.go
```diff
@@ -109,6 +110,11 @@ func main() {
 
 	k8sClient := k8s.NewClient(sharedInformers)
 
+	var proxyAddr string
+	if proxyAddrOverride != nil {
+		proxyAddr = *proxyAddrOverride
+	}
+
```
siggy, line 117: since `k8sClient` needs `proxyAddr`, how about set it in `k8s.NewClient()`, and avoid passing it through `linkerdInfoHandler`.

siggy (commented):
looking good! left a comment around proxyAddrOverride. going to test this against the latest https://github.com/BuoyantIO/bcloud/pull/3292 and then give this a final review.

agent/main.go
zaharidichev: that was my initial approach but it seemed a bit too specific for a k8s client. But yes, lets do that

siggy (approved):
tested against https://github.com/BuoyantIO/bcloud/pull/3292 and all looks good! shipit pending my last `proxyAddrOverride` comment. 🚢 👍

PR #17, opened 2021-07-01
Add Helm chart

### LINKED ISSUE #38: Helm chart for the agent
It would be great to have a helm chart for the agent to support general GitOps Workflows and helm goodness.

alpeb:
Fixes #38

Initial stab at the Helm chart for the linkerd-buoyant extension. The
required parameters in `values.yaml` are `agentID`, `agentKey`,
`agentDownloadKey` and `agentClusterName`. Please correct if those are not
right.

### How to test:
```
$ helm install linkerd-buoyant --set agentClusterName=foobar --set agentID=abc --set agentKey=xyz --set agentDownloadKey=rst chart/linkerd-buoyant

NAME: linkerd-buoyant
LAST DEPLOYED: Thu Jul  1 17:41:48 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Thank you for installing linkerd-buoyant.

Your release is named linkerd-buoyant.

To help you manage your Buoyant Cloud agent, you can install the CLI extension by running:

  curl -sL https://buoyant.cloud/install | sh

Alternatively, you can download the CLI directly via the linkerd-buoyant releases page:

  https://github.com/BuoyantIO/linkerd-buoyant/releases

To make sure everything works as expected, run the following:

  linkerd-buoyant check

Looking for more? Visit https://buoyant.io/cloud/
```

The `README.md` was generated with `bin/helm-docs` from the linkerd2
repo.

People can use this chart as it stands in this repo using the command
above, but we should publish it in a Helm repo (linkerd's Helm repo?) to
make it visible and searchable in artifacthub.io

As mentioned in the issue referred above, it remains to hook this into
CI (either here or bcloud's repo) to automatically update and publish
this chart.


klingerf:
@alpeb Thank you so much for putting this together! It's going to be a huge improvement. I've added myself and @zaharidichev as reviewers, and we'll give you some feedback in the next day or two. Thanks again for working on this.

chart/linkerd-buoyant/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| accountDownloadKey | string | `nil` | The download key for your Buoyant Cloud account. Required. |
+| accountID | string | `nil` | The ID for your Buoyant Cloud account. Required. |
+| accountKey | string | `nil` | The key for your Buoyant Cloud account. Required. |
+| agent.logLevel | string | `"debug"` | Log level for the buoyant-cloud-agent pod |
+| clusterName | string | `nil` | The name of your cluster, as it should appear in Buoyant Cloud. Required |
+| installNamespace | bool | `true` | Set to false when installing Linkerd in a custom namespace. If so, make sure you add the required annotations and labels, as depicted in the namespace.yaml template |
```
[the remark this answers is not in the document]
zaharidichev: Infact I kind of take that back. For local developement, we install a docker-host manifest so we can connect to the docker-compose env from the kind cluster, so we already have a namespace. So yes, this flag is needed actually

chart/linkerd-buoyant/templates/agent.yaml
```diff
@@ -0,0 +1,438 @@
...
+    spec:
+      securityContext:
+        fsGroup: 1000
+      serviceAccount: buoyant-cloud-agent
+      containers:
+      - name: buoyant-cloud-agent
+        image: ghcr.io/buoyantio/linkerd-buoyant:v0.2.0
+        imagePullPolicy: Always
+        args:
+        - "-grpc-addr=api.staging-buoyant.cloud:443"
```
zaharidichev, line 56: I am pretty sure this needs to be configurable and the default should probably point to prod. Also, we need to be able to change that for local developement to point to something like `dockerhost:8086`

chart/linkerd-buoyant/templates/agent.yaml
```diff
@@ -0,0 +1,438 @@
...
+        - "-log-level={{ .Values.agent.logLevel }}"
```
zaharidichev, line 57: There is a `-insecure` option on the agent that disables TLS. Again, I think we would need that being here so we can run the whole env in dev

chart/linkerd-buoyant/templates/agent.yaml
```diff
@@ -0,0 +1,438 @@
...
+        scrape_interval: 10s
+        external_labels:
+          cluster_id: ${BUOYANT_CLOUD_ID}
+          cluster_name: ${BUOYANT_CLOUD_NAME}
+      configs:
+      - host_filter: false
+        name: buoyant-cloud-metrics
+        wal_truncate_frequency: "1m"
+        remote_write:
+        - url: https://api.staging-buoyant.cloud:443/remote-write
```
zaharidichev, line 123: Configurable/defaulting to prod ?

zaharidichev (commented):
I think this is great. And thanks a bunch for doing this tedious work... I left a few comments. One thing that might be good to think about.. Currently the install flow uses a yaml template defined in the bloud repo to output the manifest. So having this, we will end up with two templates that we need to keep in sync. Is there any clever way of centralizing that. I know we can use a similar approach to Linkerd where we use a helm chart for both CLI (in the case of BC a web install) and Helm installations. Can we do something similar. But then where would that chart live. Can we make it so the helm chart lives here in this repo and Bcloud depends on it somehow? Just thinking outloud.

chart/linkerd-buoyant/README.md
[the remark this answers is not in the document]
alpeb: Oops, I'll fix that "linkerd" reference to "linkerd-buoyant agent". More context about this flag: turns out since Helm v3 the best practice is to actually have the templates be namespace-agnostic, and set it with the `--namespace` flag in `helm install`. Mainly because the Helm secret holding the config will live under that same namespace. If we hard-code the namespace then `helm install` will unfortunately put that secret in the default namespace, which is not desirable for obvious reasons.

BUT given we require specific metadata on this namespace, it's just easier for folks if we create it, leaving the option for them to manage it themselves if they're concerned about the secret location.

chart/linkerd-buoyant/README.md
```diff
@@ -0,0 +1,37 @@
...
+Kubernetes: `>=1.15.0-0`
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| accountDownloadKey | string | `nil` | The download key for your Buoyant Cloud account. Required. |
+| accountID | string | `nil` | The ID for your Buoyant Cloud account. Required. |
+| accountKey | string | `nil` | The key for your Buoyant Cloud account. Required. |
+| agent.logLevel | string | `"debug"` | Log level for the buoyant-cloud-agent pod |
```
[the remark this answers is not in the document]
alpeb: Just being future-proof by creating `agent` and `metrics` sections in `values.yaml` to group together configs that we might wanna add later on to further customize those two deployments.

chart/linkerd-buoyant/templates/agent.yaml
```diff
@@ -0,0 +1,438 @@
...
+  name: buoyant-cloud-id
+  namespace: {{ .Values.namespace }}
+  labels:
+    app.kubernetes.io/part-of: buoyant-cloud
+type: Opaque
+data:
+  id: {{ b64enc (required "Please provide the ID for your Buoyant Cloud account in the accountID value" .Values.accountID | trim) }}
+  key: {{ b64enc (required "Please provide the key for your Buoyant Cloud account in the accountKey value" .Values.accountKey | trim) }}
+  downloadKey: {{ b64enc (required "Please provide the download key for your Buoyant Cloud account in the accountDownloadKey value" .Values.accountDownloadKey | trim) }}
+  name: {{ b64enc (required "Please provide the cluster named the clusterName value" .Values.clusterName | trim) }}
```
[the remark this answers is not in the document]
alpeb: :+1:

chart/linkerd-buoyant/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+# -- The download key for your Buoyant Cloud account. Required.
+accountDownloadKey:
+
+# -- The name of your cluster, as it should appear in Buoyant Cloud. Required
+clusterName:
+
+# -- Namespace where the agent components will be deployed
+namespace: buoyant-cloud
+
+# -- Set to false when installing Linkerd in a custom namespace. If so, make
```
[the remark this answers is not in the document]
alpeb: :+1:

chart/linkerd-buoyant/templates/agent.yaml
alpeb: :+1:

chart/linkerd-buoyant/templates/agent.yaml
alpeb: :+1:

chart/linkerd-buoyant/templates/agent.yaml
alpeb: :+1:

alpeb:
Thanks for the great feedback! I've just pushed an update.

I agree with your point about avoiding having to maintain two copies of the template. Given the agent is open source, the template might be better located here? And so we can have here a github action that publishes the chart.
And like you say,  `linkerd-buoyant install` could render it through the Helm lib, just calling the bcloud API to retrieve params.... 

I guess the only need for this template on bcloud's side is for downloading it from the UI. So the code doing the rendering in this repo could be modularized in a go package that can be imported in bcloud for fulfilling that purpose. WDYT?

chart/linkerd-buoyant/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
+name: linkerd-buoyant
+version: 0.2.0
```
siggy, line 3: can the version be templatized? currently we don't hard-code the `linkerd-buoyant` version anywhere in the repo.

chart/linkerd-buoyant/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+# -- The name of your cluster, as it should appear in Buoyant Cloud. Required
+clusterName:
+
+# -- Namespace where the agent components will be deployed
+namespace: buoyant-cloud
+
+# -- Set to false when installing linkerd-buoyant in a custom namespace. If so,
+# make sure you add the required annotations and labels, as depicted in the
+# namespace.yaml template
+installNamespace: true
```
siggy, line 19: any reason not to just hard-code this? we currently don't support running the agent in different namespaces.

siggy (commented):
thanks for doing this! i left some minor comments, they may not make sense as i'm not that familiar with helm. on that note, some comments:

- today, `linkerd-buoyant install` retrieves a fully-formed agent manifest yaml from the bcloud server. this ensures that the manifest and image tags are for the latest version. is this going to change? how does helm integration interact with this?
- related, the "latest version" for the agent is governed by the bcloud server. this allows us to publish new docker images in this repo, test them, and then later on, update bcloud server to reference the new image tags in the manifest yamls. this could motivate keeping helm publishing in the bcloud repo, unless it's useful to keep that part open source as well.
- long-term we probably want the agent manifest template in one place, but maybe not a blocker for this PR.

chart/linkerd-buoyant/values.yaml
zaharidichev: Its a niche usecase, but if we want to use the chart locally, we would want to opt out out outputting the namespace manifest as we should already have a namespace because of the dockerhost config being applied beforehand. Right?

alpeb: More importantly, when people in prod are given the namespace they can work in, and don't have access to (or don't want to write to) the `default` namespace, which is where Helm v3 will put its config secret if its not instructed to use an existing namespace (through `helm install --namespace xxx`)

alpeb:
> * today, `linkerd-buoyant install` retrieves a fully-formed agent manifest yaml from the bcloud server. this ensures that the manifest and image tags are for the latest version. is this going to change? how does helm integration interact with this?

The helm workflow (at least without getting into hooks which IMO is not granted here) can only render the templates declared in the chart, not fetched through the network. The id and keys would need to be fed as parameters to `helm install` so bcloud's UI (or maybe this very same linkerd-buoyant extension) would need to show them somehow to the user.

The CLI workflow, instead of retrieving the whole manifest from bcloud's API, would just need to retrieve the keys. And then using the newish go's `embed` package and the helm lib it can render the templates, similar to what we do for linkerd's CLI.

> * related, the "latest version" for the agent is governed by the bcloud server. this allows us to publish new docker images in this repo, test them, and then later on, update bcloud server to reference the new image tags in the manifest yamls. this could motivate keeping helm publishing in the bcloud repo, unless it's useful to keep that part open source as well.
> * long-term we probably want the agent manifest template in one place, but maybe not a blocker for this PR.

Now the only use case left for bcloud's need for those manifests is when the user downloads them manually from the UI, correct?
To avoid having a copy of the manifests at bcloud and here, and given the bcloud codebase has access to the this repo, but not the other way around, my proposal was to have the `linkerd-buoyant` repo be the SSOT for those manifests formatted as helm templates, but exposed as go packages (again, using `embed`). So bcloud's code would import them and similar to how the `linkerd-buoyant` CLI would do, it would render them using the helm lib. If you find this scheme suitable, I'd be glad to lend a hand to implement this in the bcloud repo.

Anyways that'd definitely be something for a separate PR. The current one will leave us with a usable chart in case anyone asks for it.

siggy (approved):
Thanks for the changes and explanations. I think we have a bit more to figure out around getting agent-specific data into a user's helm setup, and also reconciling the agent manifest into a SSOT, but I think we can address all those in follow up PRs. For now it's helpful to just have this Helm chart in place for testing and iterating.

zaharidichev (approved):
LGTM

siggy:
@alpeb Let's hold off on merging this until we've got a plan for how it integrates with bcloud.

.github/workflows/actions.yml
```diff
@@ -127,10 +127,62 @@ jobs:
...
+      with:
+        python-version: 3.7
+
+    - name: Set up chart-testing
+      uses: helm/chart-testing-action@v2.1.0
+
+    - name: Validate chart README.md is up to date
+      run: |
+        bin/helm-docs --chart-search-root=charts
+        git diff --exit-code charts/linkerd-buoyant/README.md
```
klingerf, line 154: TIL `--exit-code` flag

klingerf (approved):
⭐  Looks great! Tested it locally and the chart is working end-to-end to me. Left some comments about the chart itself in the other PR, which I think will make sense to address here instead. Wasn't able to test the new github actions, but am assuming it all works. Thanks for reviving this one!

PR #19, opened 2021-07-08
Add agent commands functionality + flare

zaharidichev:
This is much of the scafolding needed to support sending
commands to the agent. To begin with, we will be sending
commands to request diagnostic data about the Linkerd proxy.
This data payload will come through the `LinkerdInfo` endpoint.


agent/pkg/handler/linkerd_info.go
```diff
@@ -75,3 +92,53 @@ func (h *LinkerdInfo) handleCertsInfo() {
...
+		err = h.api.ProxyDiagnostics(m)
+		if err != nil {
+			h.log.Errorf("error sending ProxyDiagnostics message: %s", err)
+		}
+
+		err = h.k8s.SetProxyLogLevel(podName, namespace, initialLogLevel)
+		if err != nil {
+			h.log.Errorf("error reverting proxy log level to %s for pod %s/%s: %s", initialLogLevel, namespace, podName, err)
+			return
+		}
```
siggy, line 142: i think we'd want this in a `defer` to ensure it always gets called.

going a step further, related to my comment around separating log level and diagnostics collection, this may be moot.

siggy (commented):
thanks for sharing this draft. i think it's going to help inform our api design, and maybe even product requirements.

zaharidichev (commented):
forgot to add some initial comments that I had put upon posting this branch

agent/pkg/k8s/proxy.go
```diff
@@ -0,0 +1,21 @@
...
+
+// GetProxyLogs retrieves the proxy logs of a pod
+func (c *Client) GetProxyLogs(podName, namespace string) (string, error) {
+	return "here are some proxy logs", nil
+}
+
+// GetPrometheusScrape retrieves the raw prom scrape from the proxy of a pod
+func (c *Client) GetPrometheusScrape(podName, namespace string) (string, error) {
+	return "here are some prom metrics", nil
+}
```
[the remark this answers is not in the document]
zaharidichev: For hitting the endpoints, I think we can get away with just the pod DNS, but if that does not work we will most likely need to get the IP addr.  I was just thinking that this is the right way to put them since we are interacting with pods that are on the cluster even if we do not directly use the k8s api. Wrt to getting the actual logs, yes I was planning to use the go client to do that.

agent/pkg/api/manage_agent.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		c.log.Info("ManageAgentClient initialized")
+
+		break
+	}
+
+	c.log.Info("ManageAgentClient connected")
+	return client
+}
```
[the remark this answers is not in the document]
zaharidichev: I agree we can simplify things a bit here. However, I am not sure we actually need a mutex. I believe that the workloads stream needs to be guarded by a mutex because it is being accessed by callbacks coming from the shared informers machinery which runs in separate go routines. So there may be concurrent calls to `sendMessage` and  `resetStream`. With the agent commands stream though we are just pulling messages from one go routine so I do not see much point in using any synchronization here. Am I missing something here?

agent/pkg/handler/manage_agent.go
```diff
@@ -0,0 +1,58 @@
...
+			}
+		}
+	}
+}
+
+// Stop terminates the handler.
+func (h *ManageAgent) Stop() {
+	h.log.Info("shutting down")
+	close(h.stopCh)
+}
```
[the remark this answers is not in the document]
zaharidichev: I share the concerns. Will rework

agent/pkg/handler/linkerd_info.go
zaharidichev: Good point

agent/main.go
```diff
@@ -148,7 +148,10 @@ func main() {
 	// create handlers
 	eventHandler := handler.NewEvent(k8sClient, apiClient)
 	workloadHandler := handler.NewWorkload(k8sClient, apiClient)
-	linkerdInfoHandler := handler.NewLinkerdInfo(k8sClient, apiClient)
+
+	diagnostics := make(chan handler.DiagnosticsDataRequest, 1)
+	linkerdInfoHandler := handler.NewLinkerdInfo(k8sClient, apiClient, diagnostics)
+	manageAgentHandler := handler.NewManageAgent(apiClient, diagnostics)
```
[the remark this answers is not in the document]
zaharidichev: You are absolutely right. Looking at the protobuf definition it not clear what actually happens when you send a message to the `ManageAgent`. Even worse, my line of thinking was that we can keep doing that for future functionality. Namely have a new command on `ManageAgent` that can do one of a few things: 

- send no response at all
- send a response to the LinkerdInfo endpoint
- send a response to any other endpoint that we do not have at the moment but we might have in the future.

This does not feel like a good design, I was mostly governed by the feedback I got on the notion doc but now I realize it makes more sense from API design perspective to have things be a bit more principled. 

So should we have a bi directional stream instead that multiplexes all agent related communication where we send a command and expect to receive a response as a result of this command? Would that make sense. At least then the protobuf api will more precisely capture the communication patterns that are possible

agent/pkg/handler/linkerd_info.go
```diff
@@ -8,32 +8,47 @@ import (
 	pb "github.com/buoyantio/linkerd-buoyant/gen/bcloud"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/encoding/prototext"
+	"gopkg.in/yaml.v2"
 )
 
 const (
-	linkerdInfoInterval = time.Minute
+	linkerdInfoInterval       = time.Minute
+	diagnosticCollectDuration = time.Minute * 5
```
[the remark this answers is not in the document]
zaharidichev: Good question and I think that kind of relates again with the overlap between this feature and the logs one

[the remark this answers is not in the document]
klingerf: I think we should wait for more guidance from the Linkerd team here. It may not be appropriate to change the log level at all, or maybe we only want to set `trace` for a few libraries, and leave the rest at debug or info? Maybe 1 minute's worth of logs will suffice?

Regarding this comment:

> perhaps giving the user a separate, "Enable trace logging" button, or, stream metrics and logs back immediately, but also continue streaming trace logs as they come in?

The ability to change the proxy log level from the UI is on the feature roadmap, and I anticipate that we'd do that via a separate message type on the `ManageAgent` stream. Likewise we'll also have the ability to view recent logs for any given proxy, but that won't reuse the log collection code that we're adding for diagnostics, since the use cases are substantially different (get n most recent logs without adjusting the log level, versus collecting as many logs as we can for diagnostics).

agent/main.go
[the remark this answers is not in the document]
klingerf: I don't think that we should use a bidirectional stream that both sends commands to the agent and receives the results of running those commands. My thinking here is that some commands may not need responses (e.g adjusting the proxy log level), whereas other commands may send us lots of data. If we use a bidirectional stream then we can't balance the responses across all of our API servers. It would be better to send the command on one stream, and receive the data via separate unary RPCs (or streaming RPCs as needed).

If we're worried about maintaining a more principled relationship between commands and responses, I suggest that we address that with consistent naming across our API definition. For example, we could provide a `ProxyDiangostics` endpoint that mirrors the naming of the `GetProxyDiagnostics` command that we're adding here. That said, I don't really mind that the current implementation is calling the `LinkerdInfo` endpoint instead.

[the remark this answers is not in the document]
zaharidichev: > If we're worried about maintaining a more principled relationship between commands and responses, I suggest that we address that with consistent naming across our API definition. For example, we could provide a ProxyDiangostics endpoint that mirrors the naming of the GetProxyDiagnostics command that we're adding here. That said, I don't really mind that the current implementation is calling the LinkerdInfo endpoint instead.

Yes, we already covered that in our Friday conversation and I have reflected that in the design doc, so this is how we proceed

agent/pkg/api/manage_agent.go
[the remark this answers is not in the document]
zaharidichev: I actually like the older approach quite a bit. At least all the logic is isolated in one place. I think I will adopt that for both streams

agent/pkg/handler/linkerd_info.go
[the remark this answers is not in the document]
zaharidichev: Alright, lets wait for input from the Linkerd team

agent/pkg/k8s/proxy.go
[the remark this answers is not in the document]
siggy: i had mostly envisioned `k8s.Client` as a client for the k8s api, but this is not a blocker. as we move forward with implementation, see where it feels most ergonomic to put a `/metrics` scrape.

proto/buoyant-cloud-api.proto
```diff
@@ -133,4 +155,5 @@ service Api {
   rpc WorkloadStream(stream WorkloadMessage) returns (Empty) {}
   rpc AddEvent(Event) returns (Empty) {}
   rpc LinkerdInfo(LinkerdMessage) returns (Empty) {}
+  rpc ManageAgent(Auth) returns (stream AgentCommand) {}
```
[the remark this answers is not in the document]
siggy: Yup `Auth` is sufficient.

In hindsight, I'm wondering if we should have just used basic auth, rather than putting an `Auth` into each API call. Something to consider longer term.

agent/pkg/handler/manage_agent.go
```diff
@@ -0,0 +1,96 @@
...
+				h.log.Errorf("Error from stream %s", err)
+				continue
+			}
+			switch command := agentCommand.Command.(type) {
+			case *pb.AgentCommand_GetProxyDiagnostics:
+				proxyDiagnostic := command.GetProxyDiagnostics
+				go h.handleProxyDiagnostics(proxyDiagnostic.PodName, proxyDiagnostic.PodNamespace, proxyDiagnostic.DiagnosticId)
+			}
+		}
+	}
```
siggy, line 50: rather than blocking on `h.api.RecvAgentCommand()`, what if that method returned a channel we can receive commands on, that way as soon as a `ManageAgent.Stop()` is called, `ManageAgent` stops trying to receive commands?

going a step further (and related to my comment around replacing `api.Client.CloseAgentCommandStream()` with a more general `api.Client.Close()`), if we can ensure that calling `ManageAgent.Stop()` will stop it from trying to receive, we should be able to safely call `api.Client.Close()` from main, and do away with the `client_closed` boolean in `manageAgentStream`.

proto/buoyant-cloud-api.proto
```diff
@@ -125,6 +125,34 @@ message LinkerdMessage {
...
+  string diagnostic_id = 1;
+  string pod_name = 2;
+  string pod_namespace = 3;
+}
+
+message AgentCommand {
+  oneof command {
+    GetProxyDiagnostics get_proxy_diagnostics = 1;
+  }
+}
```
siggy, line 142: i like this approach 👍 . naming the `AgentCommand.GetProxyDiagnostics` to match the `ProxyDiagnostics` rpc satisfies my primary concern around making explicit which commands map to which messages/apis.

...re: @klingerf's comment:
> If we're worried about maintaining a more principled relationship between commands and responses, I suggest that we address that with consistent naming across our API definition. For example, we could provide a ProxyDiangostics endpoint that mirrors the naming of the GetProxyDiagnostics command that we're adding here. That said, I don't really mind that the current implementation is calling the LinkerdInfo endpoint instead.

adding `ProxyDiagnostic` to `LinkerdMessage` `oneof` would also satisfy my primary concern. i realize we built `LinkerdMessage` with a `oneof` to be extensible, but i have a slight preference for the separate `ProxyDiagnostics` api as the alternative feels like overloading `LinkerdInfo`. also implementing `ProxyDiagnostics` as a separate api makes the implementation clearer.

ultimately you're (@zaharidichev) the one who will spend the majority of the time implementing/testing/maintaining this (at least in the short term), so i'll defer to you on the approach here.

siggy (commented):
thanks for updates!

agent/pkg/handler/manage_agent.go
zaharidichev: That sounds a lot more elegant. However we still need to put some async code in a goroutine to fill this chan. I assume this would live in the `manage_agent_stream.go`

agent/main.go
```diff
@@ -41,8 +42,8 @@ func main() {
 	grpcAddr := flag.String("grpc-addr", "api.buoyant.cloud:443", "address of the Buoyant Cloud API")
 	kubeConfigPath := flag.String("kubeconfig", "", "path to kube config")
 	logLevel := flag.String("log-level", "info", "log level, must be one of: panic, fatal, error, warn, info, debug, trace")
+	localMode := flag.Bool("local-mode", false, "enable port forwarding for local mode")
```
siggy, line 45: `enable port forwarding for local development` to be even more explicit.

agent/pkg/handler/manage_agent.go
```diff
@@ -0,0 +1,102 @@
...
+// Stop terminates the handler.
+func (h *ManageAgent) Stop() {
+	h.log.Info("shutting down")
+	close(h.stopCh)
+}
+
+func (h *ManageAgent) handleProxyDiagnostics(ctx context.Context, podName, namespace, diagnosticID string) {
+	logs, err := h.k8s.GetProxyLogs(ctx, podName, namespace)
+	if err != nil {
+		h.log.Errorf("cannot obtain proxy logs for diagnosticID: %s", diagnosticID)
```
siggy, line 58: log the `error` as well.

agent/pkg/handler/manage_agent.go
```diff
@@ -0,0 +1,102 @@
...
+		h.log.Errorf("cannot obtain nodes for diagnosticID: %s", diagnosticID)
+	} else {
+		h.log.Debugf("Successfully obtained nodes for diagnosticID: %s", diagnosticID)
+	}
+
+	svcManifest, err := h.k8s.GetK8sServiceManifest(ctx)
+	if err != nil {
+		h.log.Errorf("cannot K8s svc manifest for diagnosticID: %s", diagnosticID)
+	} else {
+		h.log.Debugf("Successfully K8s svc manifest for diagnosticID: %s", diagnosticID)
```
siggy, line 95: `Successfully obtained K8s`

agent/pkg/handler/manage_agent.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	err = h.api.ProxyDiagnostics(diagnosticID, logs, metrics, podSpec, ld5ConfigMap, nodes, svcManifest)
+	if err != nil {
+		h.log.Debugf("error sending ProxyDiagnostics message: %s", err)
```
siggy, line 100: `h.log.Errorf`

agent/pkg/k8s/k8s.go
```diff
@@ -54,7 +60,7 @@ const (
 
 var errSyncCache = errors.New("failed to sync caches")
 
-func NewClient(sharedInformers informers.SharedInformerFactory, proxyAddrOverride string) *Client {
+func NewClient(k8sClient kubernetes.Interface, sharedInformers informers.SharedInformerFactory, k8sConfig *rest.Config, ld5API *ld5k8s.KubernetesAPI) *Client {
```
siggy, line 63: `k8sConfig` is unused

siggy, line 63: `k8sConfig` is unused.

agent/pkg/handler/manage_agent.go
```diff
@@ -0,0 +1,102 @@
...
+	} else {
+		h.log.Debugf("Successfully obtained pod manifest for diagnosticID: %s", diagnosticID)
+	}
+
+	metrics, err := h.k8s.GetPrometheusScrape(ctx, podName, namespace)
+	if err != nil {
+		h.log.Errorf("cannot obtain proxy metrics for diagnosticID: %s", diagnosticID)
+	} else {
+		h.log.Debugf("Successfully obtained proxy metrics for diagnosticID: %s", diagnosticID)
+	}
```
siggy, line 75: not for this PR, but we could consider running all these operations concurrently to speed things up, though i assume other than `GetPrometheusScrape` taking 1 minute, everything else is pretty fast.

agent/pkg/k8s/proxy.go
```diff
@@ -0,0 +1,159 @@
...
+				p := p
+				port = &p.ContainerPort
+				break
+			}
+		}
+		if port == nil {
+			return nil, fmt.Errorf("cannot find proxy admin port for pod: %s/%s", namespace, podName)
+		}
+		metricsUrl = fmt.Sprintf("http://%s:%d/metrics", pod.Status.PodIP, *port)
+	}
```
siggy, line 93: similar comment to above, consider moving into a helper.

agent/pkg/k8s/proxy.go
```diff
@@ -0,0 +1,159 @@
...
+			break
+		}
+		time.Sleep(time.Second * 10)
+	}
+
+	return data, nil
+}
+
+// GetPodManifest retrieves pod manifest
+func (c *Client) GetPodSpec(ctx context.Context, podName, namespace string) (*pb.Pod, error) {
```
siggy, line 118: GetPodManifest vs. GetPodSpec

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	google.golang.org/grpc v1.39.0
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0 // indirect
 	google.golang.org/protobuf v1.27.1
+	gopkg.in/yaml.v2 v2.4.0
```
siggy, line 14: intentional? i don't see it used in the code.

siggy (commented):
(apologies i hit submit prematurely on that last set of comments)

it all works!
```
DEBU[2021-07-16T22:58:34Z] Successfully obtained proxy logs for diagnosticID: z3VJ8P7Tc22wHdaD  handler=manage_agent
DEBU[2021-07-16T22:58:35Z] Successfully obtained pod manifest for diagnosticID: z3VJ8P7Tc22wHdaD  handler=manage_agent
DEBU[2021-07-16T22:58:35Z] Starting port forward to https://127.0.0.1:39737/api/v1/namespaces/emojivoto/pods/web-0/portforward 39483:4191
DEBU[2021-07-16T22:58:35Z] Port forward initialised
DEBU[2021-07-16T22:59:04Z] Starting port forward to https://127.0.0.1:39737/api/v1/namespaces/linkerd/pods/linkerd-identity-6dd85cb4f9-2mht6/portforward 37537:4191
DEBU[2021-07-16T22:59:04Z] Port forward initialised
DEBU[2021-07-16T22:59:25Z] Successfully obtained proxy metrics for diagnosticID: z3VJ8P7Tc22wHdaD  handler=manage_agent
DEBU[2021-07-16T22:59:25Z] Successfully obtained Linkerd config map for for diagnosticID: z3VJ8P7Tc22wHdaD  handler=manage_agent
DEBU[2021-07-16T22:59:25Z] Successfully obtained nodes for diagnosticID: z3VJ8P7Tc22wHdaD  handler=manage_agent
DEBU[2021-07-16T22:59:25Z] Successfully K8s svc manifest for diagnosticID: z3VJ8P7Tc22wHdaD  handler=manage_agent
```

i left some minor comments but overall everything is looking good.

agent/pkg/handler/manage_agent.go
zaharidichev: Yes I was thinking about that and reached a similar conclusion. Not sure it will make much of a difference really.

agent/pkg/k8s/proxy.go
```diff
@@ -0,0 +1,159 @@
...
+	for _, c := range pod.Spec.Containers {
+		if c.Name == ldConsts.ProxyContainerName {
+			c := c
+			proxyContainer = &c
+			break
+		}
+	}
+	if proxyContainer == nil {
+		return nil, fmt.Errorf("cannot find proxy container for pod: %s/%s", namespace, podName)
+	}
```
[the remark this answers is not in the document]
zaharidichev: good call

agent/pkg/k8s/certificates.go
```diff
@@ -137,26 +138,45 @@ func extractRootsCerts(container *v1.Container) ([]*pb.CertData, error) {
...
+		}
+
+		defer pf.Stop()
+	} else {
+		port, err := getProxyAdminPort(container)
+		if err != nil {
+			return nil, err
+		}
+
+		podAddr = fmt.Sprintf("%s:%d", pod.Status.PodIP, port)
```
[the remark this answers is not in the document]
zaharidichev: yes, made it a bit more general purpose. We will most likely needed in the future for arbitrary containers/ports

siggy (commented):
all the updates look good. i ran into one issue with the dev setup...

i tried running the agent with `-local-mode`, and without the port-forward for certs, and i'm getting an error:

```bash
$ bin/go-run agent -grpc-addr localhost:8086 --insecure -bcloud-id $AGENT_ID -bcloud-key $AGENT_KEY -log-level debug -kubeconfig ~/.kube/config localhost -local-mode
...
ERRO[2021-07-19T18:11:47Z] error getting control plane certs: dial tcp 10.244.0.11:4191: i/o timeout  handler=linkerd_info
```

...similar issue when getting diagnostics:
```
DEBU[2021-07-19T18:17:01Z] Successfully obtained proxy logs for diagnosticID: eF4kNm9G8PhSSQjQ  handler=manage_agent
DEBU[2021-07-19T18:17:01Z] Successfully obtained pod manifest for diagnosticID: eF4kNm9G8PhSSQjQ  handler=manage_agent
ERRO[2021-07-19T18:17:31Z] cannot obtain proxy metrics for diagnosticID eF4kNm9G8PhSSQjQ: Get "http://10.244.0.10:4191/metrics": dial tcp 10.244.0.10:4191: i/o timeout  handler=manage_agent
DEBU[2021-07-19T18:17:31Z] Successfully obtained Linkerd config map for for diagnosticID: eF4kNm9G8PhSSQjQ  handler=manage_agent
DEBU[2021-07-19T18:17:31Z] Successfully obtained nodes for diagnosticID: eF4kNm9G8PhSSQjQ  handler=manage_agent
DEBU[2021-07-19T18:17:31Z] Successfully obtained K88s svc manifest for diagnosticID: eF4kNm9G8PhSSQjQ  handler=manage_agent
```

i've confirmed i can access http://10.244.0.10:4191/metrics from inside my kind cluster. even stranger, running `kubectl -n linkerd port-forward deploy/linkerd-identity 4191` did not fix it (also confirmed that with port-forwarding, i could access localhost:4191/metrics from my vm, outside k8s)

siggy (approved):
great work! 👍 🚢 

confirmed `-local-mode` worked for me this time, something must have been wrong with my dev env previously:
```
DEBU[2021-07-20T19:09:39Z] Starting port forward to https://127.0.0.1:39737/api/v1/namespaces/linkerd/pods/linkerd-identity-6dd85cb4f9-2mht6/portforward 33287:4191
DEBU[2021-07-20T19:09:39Z] Port forward initialised
```

agent/pkg/api/test_helpers.go
```diff
@@ -72,6 +108,24 @@ func (m *MockBcloudClient) LinkerdMessages() []*pb.LinkerdMessage {
...
+	m.Lock()
+	defer m.Unlock()
+
+	messages := make([]*pb.ProxyDiagnostic, len(m.proxyDiagnosticMessages))
+	for i, m := range m.proxyDiagnosticMessages {
+		messages[i] = m
+	}
+
+	return messages
+}
```
klingerf, line 121: Just curious -- is it necessary to create a new slice here? Why not just return `m.proxyDiagnosticMessages`? Same goes for the other *Messages() functions above. It seems like they could all be simplified to:

```go
func (m *MockBcloudClient) ProxyDiagnosticMessages() []*pb.ProxyDiagnostic {
	m.Lock()
	defer m.Unlock()
	return m.proxyDiagnosticMessages
}
```

klingerf (approved):
⭐  Awesome! I was able to successfully test this end-to-end.

```
Successfully obtained proxy logs for diagnosticID: 6tvpLeBBMDhI44Bp  handler=manage_agent
Successfully obtained pod manifest for diagnosticID: 6tvpLeBBMDhI44Bp  handler=manage_agent
Successfully obtained proxy metrics for diagnosticID: 6tvpLeBBMDhI44Bp  handler=manage_agent
Successfully obtained Linkerd config map for for diagnosticID: 6tvpLeBBMDhI44Bp  handler=manage_agent
Successfully obtained nodes for diagnosticID: 6tvpLeBBMDhI44Bp  handler=manage_agent
Successfully obtained K8s svc manifest for diagnosticID: 6tvpLeBBMDhI44Bp  handler=manage_agent
```

This implementation is a whole lot cleaner and clearer from where it started. Nice work! I left a few naming / organizational comments below. I do think we should rename all occurrences of `ld5` to `l5d`, since that's the more common shorthand for linkerd.

agent/pkg/k8s/certificates.go
```diff
@@ -137,26 +126,22 @@ func extractRootsCerts(container *v1.Container) ([]*pb.CertData, error) {
...
-	if proxyAddrOverride != "" {
-		podAddr = proxyAddrOverride
-	}
+	defer proxyConnection.cleanup()
 
 	conn, err := tls.DialWithDialer(
 		&net.Dialer{Timeout: 5 * time.Second},
 		"tcp",
-		fmt.Sprintf("%s:%d", podAddr, port), &tls.Config{
+		proxyConnection.host, &tls.Config{
```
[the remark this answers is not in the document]
zaharidichev: Oh..... that certainly look ugly. Should have noticed it.

agent/pkg/api/manage_agent_stream.go
```diff
@@ -0,0 +1,83 @@
...
+	// protects stream
+	sync.Mutex
+}
+
+func newManageAgentStream(auth *pb.Auth, client pb.ApiClient) *manageAgentStream {
+	return &manageAgentStream{
+		auth:     auth,
+		client:   client,
+		log:      log.WithField("stream", "ManageAgentStream"),
+		commands: make(chan *pb.AgentCommand, 10),
```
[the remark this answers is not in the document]
zaharidichev: You are right. I got confused. We do not need a buffered one.

agent/pkg/api/test_helpers.go
zaharidichev: I was under the impression that passing a pointer to an array like this would make it unsafe because one can modify this data structure in another go routing and trigger a data race. Cannot seem to cause that to happen 🤷

klingerf (approved):
⭐  Looks good, thanks for making those updates!

agent/pkg/api/test_helpers.go
zaharidichev: Thanks for the details. That makes a lot of sense

Issue #21, opened 2021-07-26
ManageAgent stream disconnects/reconnects


siggy:
The `ManageAgent` server streaming API disconnects/reconnects every 10 seconds:
```
time="2021-07-26T19:44:20Z" level=info msg="ManageAgentStream connected" stream=ManageAgentStream
time="2021-07-26T19:44:30Z" level=info msg="stream closed, reseting: rpc error: code = Unavailable desc = connection closed" stream=ManageAgentStream
time="2021-07-26T19:44:30Z" level=info msg="ManageAgentStream connected" stream=ManageAgentStream
```

Relates to:
- https://github.com/linkerd/linkerd2-proxy/pull/1149
- https://github.com/linkerd/linkerd2/issues/5988

PR #22, opened 2021-07-26
Turn down some ManageAgent logs from info to debug

siggy:
The `ManageAgent` stream reconnects every 10 seconds, due to #21. While
the API will still work, it adds noise to the log.

Modify some ManageAgent logs from info to debug, until #21 is resolved.


zaharidichev (approved):
LGTM. Sorry I should have toned that down

PR #24, opened 2021-08-03
Remove proxy log level from proto schema

zaharidichev:
This PR introduces a change that removes the proxy log level 
from the `ProxyLog` proto message as there is not an ergonomic
way to get that information at the moment.


klingerf (approved):
⭐  Nice cleanup!

PR #25, opened 2021-08-05
Extract CP roots from config map

zaharidichev:
When running with the latest Linkerd edge bloud agent is sending a certificate message with the Roots being empty. The reason for that is that in the bcloud agent, we look for the `EnvVar.Value` field rather than the `EnvVar.ValueFrom`. In https://github.com/linkerd/linkerd2/pull/6455 the proxy manifest was changed to load the trust anchors from a config map. This causes the described above problem. 

This can be fixed by making the agent consider this value when inspecting the env vars. It is important to make this backwards compatible - check the `Value` field and only if not present, check the `ValueFrom` one. This might require RBAC changes.

Note to reviewers: 

In order to test, run this agent with the latest stable to ensure the change is backwards compatible. After that you can run it with the latest edge to convince yourself that the problem has been fixed. In both cases you need to make sure that the certificates in the control plane page are populated.

### DIFF
--- a/agent/pkg/k8s/certificates.go
+++ b/agent/pkg/k8s/certificates.go
@@ -3,7 +3,6 @@ package k8s
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"fmt"
 	"net"
 	"time"
@@ -106,17 +105,11 @@ func (c *Client) extractRootsCerts(ctx context.Context, container *v1.Container,
 			continue
 		}
 
-		var certificates []*x509.Certificate
-		var err error
-		if ev.Value != "" {
-			certificates, err = ldTls.DecodePEMCertificates(ev.Value)
-			if err != nil {
-				return nil, err
-			}
-		} else {
+		rootsValue := ev.Value
+		if rootsValue == "" {
 			// in this case we need to check for a reference to a config map
 			if ev.ValueFrom == nil || ev.ValueFrom.ConfigMapKeyRef == nil {
-				return nil, fmt.Errorf("neither a Value nor a ConfigMapKeyRef for the %s env var arepresent on proxy container [%s]", identity.EnvTrustAnchors, container.Name)
+				return nil, fmt.Errorf("neither a Value nor a ConfigMapKeyRef for the %s env var are present on proxy container [%s]", identity.EnvTrustAnchors, container.Name)
 			}
 			cmName := ev.ValueFrom.ConfigMapKeyRef.Name
 			cm, err := c.k8sClient.CoreV1().ConfigMaps(namespace).Get(ctx, cmName, metav1.GetOptions{})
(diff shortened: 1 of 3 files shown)


agent/pkg/k8s/certificates.go
```diff
@@ -95,15 +100,41 @@ func getServerName(podsa string, podns string, container *v1.Container) (string,
...
+		var err error
+		if ev.Value != "" {
+			certificates, err = ldTls.DecodePEMCertificates(ev.Value)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			// in this case we need to check for a reference to a config map
+			if ev.ValueFrom == nil || ev.ValueFrom.ConfigMapKeyRef == nil {
+				return nil, fmt.Errorf("neither a Value nor a ConfigMapKeyRef for the %s env var arepresent on proxy container [%s]", identity.EnvTrustAnchors, container.Name)
```
siggy, line 119: `are present`

agent/pkg/k8s/certificates.go
```diff
@@ -95,15 +100,41 @@ func getServerName(podsa string, podns string, container *v1.Container) (string,
...
+		var certificates []*x509.Certificate
+		var err error
+		if ev.Value != "" {
+			certificates, err = ldTls.DecodePEMCertificates(ev.Value)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			// in this case we need to check for a reference to a config map
+			if ev.ValueFrom == nil || ev.ValueFrom.ConfigMapKeyRef == nil {
```
siggy, line 118: slightly simpler:
```go
rootsValue := ev.Value

if rootsValue == "" {
	// in this case we need to check for a reference to a config map
	if ev.ValueFrom == nil || ev.ValueFrom.ConfigMapKeyRef == nil {
		return nil, fmt.Errorf("neither a Value nor a ConfigMapKeyRef for the %s env var are present on proxy container [%s]", identity.EnvTrustAnchors, container.Name)
	}
	cmName := ev.ValueFrom.ConfigMapKeyRef.Name
	cm, err := c.k8sClient.CoreV1().ConfigMaps(namespace).Get(ctx, cmName, metav1.GetOptions{})
	if err != nil {
		return nil, fmt.Errorf("cannot obtain config map %s/%s", namespace, cmName)
	}

	var ok bool
	rootsValue, ok = cm.Data[trustRootsConfigMapKeyName]
	if !ok {
		return nil, fmt.Errorf("config map %s/%s does not have %s key", namespace, cmName, trustRootsConfigMapKeyName)
	}
}

certificates, err := ldTls.DecodePEMCertificates(rootsValue)
if err != nil {
	return nil, err
}
```

agent/pkg/k8s/certificates_test.go
```diff
@@ -317,11 +318,12 @@ AiAtuoI5XuCtrGVRzSmRTl2ra28aV9MyTU7d5qnTAFHKSgIgRKCvluOSgA5O21p5
 	fixtures := []*struct {
 		testName      string
 		container     *v1.Container
+		crtConfigMap  *v1.ConfigMap
```
siggy, line 321: 👍

agent/pkg/k8s/certificates.go
```diff
@@ -11,16 +13,19 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
 const (
 	identityComponentName        = "identity"
 	linkerdNsEnvVarName          = "_l5d_ns"
 	linkerdTrustDomainEnvVarName = "_l5d_trustdomain"
+	trustRootsConfigMapName      = "linkerd-identity-trust-roots"
+	trustRootsConfigMapKeyName   = "ca-bundle.crt"
```
siggy, line 25: are these easily available in the `linkerd/linkerd2` module?

siggy (approved):
nice work! confirmed i could get cert info for both `stable-2.10.2` and `edge-21.7.5`. shipit provided all my comments make sense. 👍 🚢

agent/pkg/k8s/certificates.go
zaharidichev: Nice, thanks !

agent/pkg/k8s/certificates.go
zaharidichev: Unfortunately, no

PR #29, opened 2021-09-07
delay sending of events

zaharidichev:
This PR introduces a delay parameter that is used to delay starting
the events informer. This is done to aleviate the situation described in
https://github.com/BuoyantIO/bcloud/issues/1382

An alternative method would be to introduce some kind of synchronization
between the two streams but I am a bit hesitant to do that.


zaharidichev:
Closing for now as it is not a priority

PR #31, opened 2021-09-10
Update cli to check for metrics agent DaemonSet

### LINKED ISSUE #30: Update `linkerd-buoyant check` for metrics agent daemonset
# Summary
The `v0.4.3` agent release moved `buoyant-cloud-metrics` from a `Deployment` to a `DaemonSet`. The `linkerd-buoyant check` command is still looking for the old `Deployment`.

# Proposal

Modify `linkerd-buoyant check` to instead check for the `buoyant-cloud-metrics` `DaemonSet`.

siggy:
The `v0.4.3` agent release modified `buoyant-cloud-metrics` to be a
DaemonSet rather than a Deployment. The `linkerd-buoyant check` command
was not updated to check for this.

Modify `linkerd-buoyant check` to check for a DaemonSet instead.

Also modify the check hint URL to point to `linkerd-buoyant`-specific
documentation.

Fixes #30

### DIFF
--- a/cli/pkg/healthcheck/healthcheck.go
+++ b/cli/pkg/healthcheck/healthcheck.go
@@ -5,11 +5,11 @@ import (
 	"fmt"
 	"net/http"
 
+	agentk8s "github.com/buoyantio/linkerd-buoyant/agent/pkg/k8s"
 	"github.com/buoyantio/linkerd-buoyant/cli/pkg/k8s"
 	"github.com/buoyantio/linkerd-buoyant/cli/pkg/version"
 	"github.com/linkerd/linkerd2/pkg/healthcheck"
 	l5dk8s "github.com/linkerd/linkerd2/pkg/k8s"
-	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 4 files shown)


cli/pkg/healthcheck/healthcheck.go
```diff
@@ -141,46 +143,86 @@ func (hc *HealthChecker) globalChecks() []healthcheck.Checker {
...
 func (hc *HealthChecker) deploymentChecks(name string) []healthcheck.Checker {
-	var deploy *appsv1.Deployment
-	var pod v1.Pod
+	getWorkloadLabels := func(ctx context.Context) (map[string]string, error) {
+		deploy, err := hc.k8s.Deployment(ctx, name)
+		if err != nil {
+			return nil, err
+		}
+		return deploy.GetLabels(), nil
+	}
```
zaharidichev, line 152: It is a bit of a pitty there is not a more generic way to handle this

zaharidichev (approved):
LGTM

cli/pkg/healthcheck/healthcheck.go
siggy: yeah agree, i went through a similar exercise. one way this could be more generic would be to add a new method to the `k8s.Client` interface:
```go
Client interface {
	Workload(ctx context.Context, kind string, name string) (runtime.Object, error)
}
```
... but that seemed like overkill, so i left it as-is.

PR #34, opened 2021-09-27
proto: add linkerd policy CRD messages

zaharidichev:
This PR includes the proto changes to enable transferring the
CRDs for Linkerd policy from the agent to bcloud


proto/buoyant-cloud-api.proto
```diff
@@ -130,10 +130,71 @@ message CertificateInfo {
...
+  bytes server = 1;
+}
+
+message ServerMessage {
+  oneof message {
+    AddServer added = 1;
+    UpdateServer updated = 2;
+    DeleteServer deleted = 3;
+  }
+}
```
siggy, line 144: this add/update/delete pattern fits nicely with the client-go informers, and serves us well with workloads.

on the other hand, it works well for workloads because it's in the context of a `WorkloadStream` API, and also has a periodic `ListWorkloads` to handle reconciliation. if we're using the unary `LinkerdMessage` API, and we miss an add, update, or delete message, we'll be out of sync.

i can think of two options:
1) Replace add/update/delete with a periodic `AuthPolicyInfo` message with all CRDs in one message:
    ```proto
    message AuthPolicyInfo {
      repeated ServerMessage server = 1;
      repeated ServerAuthorizationMessage server_authorization = 2;
    }
    ```
2) Introduce a `AuthPolicyStream` API

i'm thinking option 1 should be sufficient, and it should minimize the increase of surface area between agent and bcloud. option 2 may be overkill. `WorkloadStream` is particularly useful to track changes during rollouts, which we don't care much for with these policy CRDs. open to other options as well.

siggy (commented):
thanks for putting this up. re-using the `LinkerdInfo` API makes sense to me. i left one comment around how we ensure bcloud stays in sync with the cluster.

proto/buoyant-cloud-api.proto
zaharidichev: That makes sense and it will be easier to implement for sure (on the agent side). I considered this approach but I had some concerns. First, `ListWorkloads` happens pretty infrequently (10 mins?). Part of the reason for that I think is the fact that we can get slammed with quite a lot of data in the List message, right? The same problem stand for the policy CRDs. Unlike the certificates objects, there could be a lot of them... potentially. So would we go for 10 minutes interval again? In that case it would be pretty non-realtime. If we go for shorter interval, say 1 minute, we risk generating tons of traffic quite frequently and be in a situation where we throw away 100% of the data most of the time, as nothing changes (like we do for certs...). Albeit not so important, me disliking this polling pattern is another reason for going with this approach. WDYT?

siggy: yeah all good points. `ListWorkloads` is ok every 10 minutes because it's _only_ for reconciliation, we're always up to date because of add/update/delete, unless we miss a message.

i had assumed we'd send policy CRDs once per minute. i also assumed that policy CRDs would be on the order of # of workloads, but are much smaller objects (that also don't include pods)? is this the case, or could we expect many more policy CRDs than workloads? also on default linkerd installs we'd expect zero?

siggy (approved):
nice and simple 👍 🚢

proto/buoyant-cloud-api.proto
```diff
@@ -130,10 +130,24 @@ message CertificateInfo {
...
+message AuthPolicyInfo {
+  repeated Server servers = 1;
+  repeated ServerAuthorization server_authorizations = 2;
+}
+
 message LinkerdMessage {
   Auth auth = 1;
   oneof message {
     CertificateInfo crt_info = 2;
+    AuthPolicyInfo auth_policy_info = 3;
```
siggy, line 150: i just remembered we'll want support for `trafficsplits.split.smi-spec.io` and `serviceprofiles.linkerd.io` as well.

siggy (commented):
we'll need serviceprofile and trafficsplit support as well (support all 4 linkerd CRDs).

proto/buoyant-cloud-api.proto
zaharidichev: that is correct, do we want to bundle all messages under one type that packs the Linkerd CRDs ?

siggy: how about we keep AuthPolicyInfo as-is, since they're quite related. then maybe add two more for the others...
```proto
  oneof message {
    CertificateInfo crt_info = 2;
    AuthPolicyInfo auth_policy_info = 3;
    ServiceProfile
    TrafficSplit
```
wdyt?

zaharidichev: Makes sense, we might also want to throw in the multicluster CRDs... Also.. I think we are facing a similar problem with these types as we are with the auth CRDs. Although we have clients for them, we do not have proto representations so we will still need to rely on JSON serialization. FWIW, all types in the k8s API have proto representations that have been generated automatically by a code-gen tool they use: https://github.com/kubernetes/api/blob/master/core/v1/generated.proto

siggy: ah, i missed the multi-cluster CRDs, i guess those come from a separate linkerd extension?

re: serializing these structs. i thought that we could instantiate serializers to handle our custom CRDs. we do related things in linkerd when converting to/from various types:
https://github.com/linkerd/linkerd2/blob/main/pkg/k8s/fake.go#L172-L193

siggy (approved):
lgtm 👍 🚢

siggy (commented):
Thinking about streaming vs. unary, i'm leaning a bit back towards unary, for a few reason:
1. `WorkloadStream` is used to detect rollouts, which lends itself well to a stateful stream of changes. We have no requirement to track changes over time of these CRDs.
2. I'm thinking we'll want to associate some CRDs with workloads. If we're only doing a full reconciliation of CRDs every 10 minutes, and we receive the CRD prior to the workload, we'll have to wait 10 minutes to properly ingest the CRD.
3. Less important, but we'll have better metrics for unary requests.

Let me know what you think.

## previous notes for posterity

~per offline convo, let's move all these new types to streaming. this approach has 3 advantages:~
1. better ux (realtime)
2. fits the informer pattern
3. less agent -> bcloud traffic~

the tradeoff is a bit more code complexity, particularly around stream and state management, but we already do this with `WorkloadStream`.

something like:
```proto
//
// LinkerdInfoStream messages
//

message Server {
  bytes server = 1;
}

message ServerAuthorization {
  bytes server_authorization = 1;
}

message TrafficSplit {
  bytes traffic_split = 1;
}

message ServiceProfile {
  bytes service_profile = 1;
}

message MulticlusterLink {
  bytes multicluster_link = 1;
}

message LinkerdInfo {
  oneof info {
    Server server = 1;
    ServerAuthorization server_authorization = 2;
    TrafficSplit traffic_split = 3;
    ServiceProfile service_profile = 4;
    MulticlusterLink multicluster_link = 5;
  }
}

message AddLinkerdInfo {
  LinkerdInfo linkerd_info = 1;
}

message DeleteLinkerdInfo {
  LinkerdInfo linkerd_info = 1;
}

message UpdateLinkerdInfo {
  LinkerdInfo old_linkerd_info = 1;
  LinkerdInfo new_linkerd_info = 2;
  google.protobuf.Timestamp timestamp = 3;
}

message ListLinkerdInfos {
  repeated LinkerdInfo linkerd_infos = 1;
}

message LinkerdStreamMessage {
  oneof message {
    Auth auth = 1;

    AddLinkerdInfo added = 2;
    UpdateLinkerdInfo updated = 3;
    DeleteLinkerdInfo deleted = 4;
    ListLinkerdInfos list = 5;
  }
}

service Api {
  rpc LinkerdInfoStream(stream LinkerdStreamMessage) returns (Empty) {}
}
```

... this can all work similarly to `WorkloadStream`, including in the `ListLinkerdInfos` message, we could include _all_ linkerd CRDs in one message, the way we send all deploys/sts/ds in one `ListWorkloads` message.

PR #35, opened 2021-09-27
Handle Linkerd CRDs 

zaharidichev:
This PR adds the logic to retrieve and send policy data to Bcloud API


agent/main.go
```diff
@@ -109,12 +110,20 @@ func main() {
...
+
 	if *localMode {
+		// in this case we can reuse the dynamic client from Linkerd
+		// as it is already there
 		l5dApi, err = l5dk8s.NewAPIForConfig(k8sConfig, "", nil, 0)
 		dieIf(err)
+		k8sDC = l5dApi.DynamicClient
+	} else {
+		k8sDC, err = dynamic.NewForConfig(k8sConfig)
+		dieIf(err)
```
siggy, line 123: ~isn't the dynamic client also available if localMode is false?~

~also now that i'm asking this, i'm trying to remember why we don't also just use `l5dApi.Interface` rather than creating our own `k8sCS`?~

ok, now i'm remembering that we instantiate `l5dApi` strictly for local mode, to enable port-forwarding. recommend always calling `dynamic.NewForConfig(k8sConfig)` ourselves, similar to how we create `k8sCS`, and add a comment explaining why we're calling `l5dk8s.NewAPIForConfig()` for local mode.

down the road, we could consider piggy-backing more on `l5dApi`, it looks like it's doing very similar things to our k8s client.

agent/pkg/handler/linkerd_info.go
```diff
@@ -45,6 +45,7 @@ func (h *LinkerdInfo) Start() {
 		select {
 		case <-ticker.C:
 			h.handleCertsInfo(context.Background())
+			h.handleAuthInfo(context.Background())
```
siggy, line 48: maybe have `policy` in the name somewhere? seeing `h.handleAuthInfo` made me think we were dealing with the `auth` proto object.

agent/pkg/k8s/policy.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"context"
+
+	pb "github.com/buoyantio/linkerd-buoyant/gen/bcloud"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// SazGVR is the GroupVersionResource for the ServerAuthorization resource.
```
siggy, line 11: can you mention this must stay in sync with https://github.com/linkerd/linkerd2/blob/22bbdad4d698cb327c4859befc3de0ddba51dfa4/pkg/k8s/policy.go#L24-L29, and also link to linkerd/linkerd2#6970 ?

going a step further, can we just import the `SazGVR` and `ServerGVR` vars from linkerd directly? (may require upgrading our linkerd2 dependency)

agent/pkg/handler/linkerd_info.go
```diff
@@ -57,6 +58,31 @@ func (h *LinkerdInfo) Stop() {
 	close(h.stopCh)
 }
 
+func (h *LinkerdInfo) handleAuthInfo(ctx context.Context) {
+	servers, err := h.k8s.GetServers(ctx)
+	if err != nil {
+		h.log.Errorf("error getting servers: %s", err)
```
siggy, line 64: if i run this with a pre-2.11 linkerd, we'll log an error every minute:
```
ERRO[2021-09-27T18:36:12Z] error getting servers: the server could not find the requested resource  handler=linkerd_info
```
is there a way to check whether the CRD is installed, or, maybe just log a warning depending on what error comes back?

agent/pkg/k8s/policy.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ServerGVR is the GroupVersionResource for the Server resource.
+var serverGVR = schema.GroupVersionResource{
+	Group:    "policy.linkerd.io",
+	Version:  "v1alpha1",
+	Resource: "servers",
+}
+
+func (c *Client) GetServers(ctx context.Context) ([]*pb.Server, error) {
+	servers, err := c.k8sDynClient.Resource(serverGVR).Namespace(metav1.NamespaceAll).List(ctx, metav1.ListOptions{})
```
siggy, line 26: TIL!

siggy (commented):
nice work! mostly minor comments.

agent/pkg/k8s/traffic_split.go
```diff
@@ -0,0 +1,35 @@
...
+	supported, err := c.resourceSupported(ts.SchemeGroupVersion.WithResource("trafficsplits"))
+	if err != nil {
+		return nil, err
+	}
+
+	if !supported {
+		return nil, nil
+	}
+
+	splits, err := c.l5dApi.TsClient.SplitV1alpha1().TrafficSplits(metav1.NamespaceAll).List(ctx, metav1.ListOptions{})
```
siggy, line 21: i see this client referencing `alpha4`, should we be using that?

siggy (commented):
thanks for updates. all minor comments.

i'm seeing `body was not decodable` in the logs, any ideas?
```
I0928 18:09:05.842300 3101009 round_trippers.go:454] GET https://127.0.0.1:40871/apis/multicluster.linkerd.io/v1alpha1?timeout=32s 404 Not Found in 1 milliseconds
I0928 18:09:05.842714 3101009 request.go:1372] body was not decodable (unable to check for Status): couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }
DEBU[2021-09-28T18:09:05Z] Resource multicluster.linkerd.io/v1alpha1, Resource=links not supported  client=k8s
```

agent/pkg/k8s/traffic_split.go
zaharidichev: Linkerd is using `SplitV1alpha1` so I think this is what we need to use

agent/pkg/k8s/k8s.go
```diff
@@ -66,14 +74,21 @@ const (
...
+	tsscheme.AddToScheme(scheme.Scheme)
+
 	protoSerializer := protobuf.NewSerializer(scheme.Scheme, scheme.Scheme)
+	yamlSerializer := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+
 	encoders := map[runtime.GroupVersioner]runtime.Encoder{
 		v1.SchemeGroupVersion:     scheme.Codecs.EncoderForVersion(protoSerializer, v1.SchemeGroupVersion),
 		appsv1.SchemeGroupVersion: scheme.Codecs.EncoderForVersion(protoSerializer, appsv1.SchemeGroupVersion),
+		ts.SchemeGroupVersion:     scheme.Codecs.EncoderForVersion(yamlSerializer, ts.SchemeGroupVersion),
+		sp.SchemeGroupVersion:     scheme.Codecs.EncoderForVersion(yamlSerializer, sp.SchemeGroupVersion),
```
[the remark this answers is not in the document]
zaharidichev: Adding it. I also unified the serialization so all these CRDs go through the `encoder` pipeline and are serialized as json

zaharidichev:
> i'm seeing `body was not decodable` in the logs, any ideas?
> ```
> I0928 18:09:05.842300 3101009 round_trippers.go:454] GET https://127.0.0.1:40871/apis/multicluster.linkerd.io/v1alpha1?timeout=32s 404 Not Found in 1 milliseconds
> I0928 18:09:05.842714 3101009 request.go:1372] body was not decodable (unable to check for Status): couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }
> DEBU[2021-09-28T18:09:05Z] Resource multicluster.linkerd.io/v1alpha1, Resource=links not supported  client=k8s
> ```

@siggy I think these are logs coming from the k8s API not being able to deal with the fact that it cannot parse the output. I think we are handling all this correctly, these are just logging statements.

```

func (c *Client) resourceSupported(gvr schema.GroupVersionResource) (bool, error) {
	if c.ignoreCRDSupportCheck {
		return true, nil
	}

	gv := gvr.GroupVersion().String()
	res, err := c.k8sClient.Discovery().ServerResourcesForGroupVersion(gv)
	if err != nil && !kerrors.IsNotFound(err) {
		return false, err
	}

	if res != nil && res.GroupVersion == gv {
		for _, apiRes := range res.APIResources {
			if apiRes.Name == gvr.Resource {
				return true, nil
			}
		}
	}

	c.log.Debugf("Resource %+v not supported", gvr)
	return false, nil
}
```

I think we are failing at ` c.k8sClient.Discovery().ServerResourcesForGroupVersion(gv)` with a NotFound and handling that correctly. WDYT?

siggy (approved):
great work!

agent/pkg/handler/linkerd_info.go
adleong: I think you can do something similar to https://github.com/linkerd/linkerd2/blob/main/pkg/k8s/authz.go#L96 to check if a CRD is present in the cluster.

PR #37, opened 2021-10-29
Use typed crds for policy

zaharidichev:
In this PR we introduce the use of the typed policy client in the agent.
Note that this uses a dev linked dependency on a commit that has not been
merged yet: https://github.com/linkerd/linkerd2/pull/7173


agent/pkg/k8s/k8s.go
```diff
@@ -92,23 +92,24 @@ func NewClient(sharedInformers informers.SharedInformerFactory, k8sClient *l5dk8
...
-	// | linkerd.io              | serviceprofiles             | spclient | json       |
-	// | split.smi-spec.io       | trafficsplits               | tsclient | json       |
-	// +-------------------------+-----------------------------+----------+------------+
+	// +-------------------------+---------------------+-----------+------------+
+	// |          Group          |        Kind         |  Client   | Serializer |
+	// +-------------------------+---------------------+-----------+------------+
+	// | policy.linkerd.io       | serverAuthorizaions | l5dClient | json       |
+	// | policy.linkerd.io       | servers             | l5dClient | json       |
+	// | multicluster.linkerd.io | links               | dynamic   | json       |
+	// | linkerd.io              | serviceprofiles     | spclient  | json       |
```
siggy, line 101: i noticed we only reference `spClient` in `main.go` when we do:
```go
spClient, err := spclient.NewForConfig(k8sConfig)
```
... and then `spClient` is passed into `k8s.NewClient()`, which takes that param as `l5dClient l5dApi.Interface`. is there another client we should instead be instantiating? feels weird to create an `spClient` and then cast it to a more generic `l5dClient`?

siggy (commented):
works great! left some comments around our various l5d clients.

agent/pkg/k8s/k8s.go
```diff
@@ -92,23 +92,24 @@ func NewClient(sharedInformers informers.SharedInformerFactory, k8sClient *l5dk8
...
+	// | split.smi-spec.io       | trafficsplits       | tsclient  | json       |
```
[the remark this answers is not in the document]
zaharidichev: We are using the `tsclient` in non test code right here: https://github.com/BuoyantIO/linkerd-buoyant/blob/zd/use-policy-crds/agent/pkg/k8s/traffic_split.go#L21. `tsclient` is not part of  `l5dClient` as `l5dClient` is coming from the codegened clients in the Linkerd repo. `tsclient` is an SMI client that is coming from another repo. It just happens so that it is being used in the `l5dk8s.KubernetesAPI{}` and this is why we access it from there. So we really have two clients that we depend on: 

- `k8sClient *l5dk8s.KubernetesAPI `- this is the Linkerd client that contains the TsClient and the Dynamic client
- `l5dClient l5dApi.Interface`  - the client that is coming from the Linkerd repo that gives us clients for all Linkerd specific CRDs (not that TS is not part of these..)

If you ask me, these can be combined into one in the Linkerd repo and that would make things cleaner but this is a separate task I think.

agent/pkg/k8s/k8s.go
zaharidichev: Right.. This is me not renaming things properly... So `spclient` is imported as: 

```
spclient "github.com/linkerd/linkerd2/controller/gen/client/clientset/versioned"
```

This is the case because before this just contained the ServiceProfile client. Now it contains the server authorization and server clients as well so it really should be: 


```
l5dApi "github.com/linkerd/linkerd2/controller/gen/client/clientset/versioned"
```

as this is not just the SP client anymore.  I have changed this. The reason we are taking an `l5dApi.Interface` instead of a `l5dApi.ClientSet` is to enable passing a fake client there, which we do in tests.

siggy (approved):
looks great works great! 🚢 👍

PR #40, opened 2021-12-28
Stop creating draft releases, fix Helm releases

siggy:
PR #17 introduced a Helm chart and related GitHub Actions to package and
release the chart. Unfortunately this caused new tag pushes to create
two GitHub releases, one draft release with our CLI assets, and one
published release with our Helm package.

Modify our GitHub Actions to always create a published release, rather
than a draft. Also replace our use of `chart-releaser-action` with
caling `chart-releaser` directly. `chart-releaser-action` always runs 1)
package 2) create release 3) update index, but we only want 1 and 3.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -261,7 +261,7 @@ jobs:
       with:
         tag_name: ${{ github.ref }}
         release_name: ${{ github.ref }}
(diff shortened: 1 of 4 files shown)


siggy:
(self-reviewing and merging to get releases working again)

PR #43, opened 2022-01-04
Replace logo in Helm chart with better icon

siggy:
We were previously using a Buoyant logo with the full company name, but
an icon is more appropriate.


zaharidichev (approved):
LGTM

PR #46, opened 2022-02-08
client credentials auth flow + autoregistration

zaharidichev:
This PR enables auto registration and client credentials auth flow.


proto/buoyant-cloud-api.proto
```diff
@@ -13,6 +13,7 @@ import "google/protobuf/timestamp.proto";
 message Empty {}
 
 message Auth {
+  option deprecated = true;
```
siggy, line 16: 👍

agent/pkg/registrator/registrator_test.go
```diff
@@ -0,0 +1,201 @@
...
+				objs = append(objs, tc.configMap)
+			}
+			cs := fake.NewSimpleClientset(objs...)
+			k8sApi := &l5dk8s.KubernetesAPI{
+				Interface: cs,
+			}
+
+			tokenRequests := 0
+			registrationRequests := 0
+			apiSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
siggy, line 119: nice!

related to comment around creating a bcloud api client, separate from `AgentRegistrator`, we could bundle this `httptest.NewServer` as part of a mock setup.

agent/pkg/registrator/registrator.go
```diff
@@ -0,0 +1,138 @@
...
+	AgentName  string `json:"agent_name"`
+	AgentID    string `json:"agent_id"`
+	IsNewAgent bool   `json:"is_new_agent"`
+}
+
+// AgentRegistrator is used to ensure that the agent that will be
+// running on the cluster is fully registered. If the agent is not
+// it will perform an automatic agent registration using the provided
+// client id and client secret credentials
+type AgentRegistrator struct {
```
siggy, line 35: `/agent/pkg/registrator.AgentRegistrator` feels a bit redundant. maybe just `Registrator`?

siggy (approved):
nice work! 🚢 👍 

since these code changes will depend on helm template changes, is the plan to keep this out of `main` until we're closer?

agent/pkg/auth/token_auth.go
```diff
@@ -0,0 +1,53 @@
...
+	tokenAddrScheme := "http"
+	if secure {
+		tokenAddrScheme = "https"
+	}
+
+	authConfig := &clientcredentials.Config{
+		ClientID:     clientID,
+		ClientSecret: clientSecret,
+		TokenURL:     fmt.Sprintf("%s://%s/agent-token?agent_id=%s", tokenAddrScheme, apiAddr, agentID),
+		AuthStyle:    oauth2.AuthStyleInHeader,
```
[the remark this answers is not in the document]
zaharidichev: Yes, this is how it uses it, it is a standard. The metrics agent performs this logic to probe for the auth style of the endpoint. If first issues a request with these in the headers and if it does not work, it issues another requests with these in the endpoint params. This is why there is an ` oauth2.AuthStyleAuto` type. We do not need to do that as we know the we expect these credentials in the basic auth header

agent/pkg/auth/token_auth.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	return &TokenPerRPCCreds{
+		secure: secure,
+		ts:     authConfig.TokenSource(context.Background()),
```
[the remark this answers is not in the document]
zaharidichev: Yes, this is true, but most people would need an http client, we need to construct an instance of `credentials.PerRPCCredentials` to use it in the GRPC clients, so an http client does not work for us here. Am I missing something?

agent/pkg/auth/token_auth.go
```diff
@@ -0,0 +1,53 @@
...
+func (t *TokenPerRPCCreds) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
+	token, err := t.ts.Token()
+	if err != nil {
+		return nil, err
+	}
+
+	return map[string]string{
+		"authorization": token.Type() + " " + token.AccessToken,
+	}, nil
+}
```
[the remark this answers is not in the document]
zaharidichev: I am not entirely sure about this but it did not work during local developement. I think that what this does is that it ensure that we have TLS in place. I am not sure we need that. Do we need it? Basically, we can add this check there and only perform it if `secure == true`, because otherwise, it errors out as the security level is not .... secure? WDYT? In any case, this errors if we run it in local mode without TLS. This is I think why by default  `RequireTransportSecurity` of their impl always returns true...  WDYT?

zaharidichev:
@siggy, yes the plan is to leave that out of main until we get the corresponding auth functionality in Bcloud merged

agent/main.go
```diff
@@ -129,20 +115,48 @@ func main() {
...
+	secure := !*insecure
+	bcloudApiClient := bcloudapi.New(id, secret, *apiAddr, secure)
+	agentRegistrator := registrator.New(bcloudApiClient, k8sAPI)
+
+	agentInfo, ts, err := agentRegistrator.EnsureRegistered(context.Background())
+	dieIf(err)
+	log.Infof("Obtained agent info: %+v", agentInfo)
+
+	// create bcloud grpc api client and streams
+	perRPCCreds := auth.NewTokenPerRPCCreds(ts, secure)
```
siggy, line 146: it feels a bit brittle that we pass `secure` into `bcloudapi.New()`, which affects the `ts` returned by `agentRegistrator.EnsureRegistered`, but then we pass both `ts` and `secure` again into `auth.NewTokenPerRPCCreds()`.

instead, how about we only set secure in `bcloudapi.New()`, move `auth` into `bcloudapi`, and get `agentRegistrator` out of the business of retrieving a `TokenSource` from `bcloudapi` and passing it back to main? and then going a step further, we can replace `AgentTokenSource` with a new method that simply returns a `credentials.PerRPCCredentials`, so the caller never has to see a `oauth2.TokenSource`. lmkwyt:

```diff
diff --git a/agent/main.go b/agent/main.go
index fa4960d..0a58404 100644
--- a/agent/main.go
+++ b/agent/main.go
@@ -10,7 +10,6 @@ import (
 	"time"

 	"github.com/buoyantio/linkerd-buoyant/agent/pkg/api"
-	"github.com/buoyantio/linkerd-buoyant/agent/pkg/auth"
 	"github.com/buoyantio/linkerd-buoyant/agent/pkg/bcloudapi"
 	"github.com/buoyantio/linkerd-buoyant/agent/pkg/handler"
 	"github.com/buoyantio/linkerd-buoyant/agent/pkg/k8s"
@@ -138,12 +137,12 @@ func main() {
 	bcloudApiClient := bcloudapi.New(id, secret, *apiAddr, secure)
 	agentRegistrator := registrator.New(bcloudApiClient, k8sAPI)

-	agentInfo, ts, err := agentRegistrator.EnsureRegistered(context.Background())
+	agentInfo, err := agentRegistrator.EnsureRegistered(context.Background())
 	dieIf(err)
 	log.Infof("Obtained agent info: %+v", agentInfo)

 	// create bcloud grpc api client and streams
-	perRPCCreds := auth.NewTokenPerRPCCreds(ts, secure)
+	perRPCCreds := bcloudApiClient.Credentials(context.Background(), agentInfo.AgentID)
 	opts := []grpc.DialOption{grpc.WithPerRPCCredentials(perRPCCreds)}
 	if *insecure {
 		opts = append(opts, grpc.WithInsecure())
diff --git a/agent/pkg/bcloudapi/client.go b/agent/pkg/bcloudapi/client.go
index f1d317d..d6098be 100644
--- a/agent/pkg/bcloudapi/client.go
+++ b/agent/pkg/bcloudapi/client.go
@@ -10,6 +10,7 @@ import (
 	"github.com/buoyantio/linkerd-buoyant/agent/pkg/k8s"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/clientcredentials"
+	"google.golang.org/grpc/credentials"
 )

 const (
@@ -28,7 +29,7 @@ type AgentInfo struct {

 type Client interface {
 	RegisterAgent(ctx context.Context, agentName string) (*AgentInfo, error)
-	AgentTokenSource(ctx context.Context, agentID string) oauth2.TokenSource
+	Credentials(ctx context.Context, agentID string) credentials.PerRPCCredentials
 }

 // ApiClient is a client that talks to the Bcloud client API
@@ -38,6 +39,7 @@ type ApiClient struct {
 	tokenAuthConfig *clientcredentials.Config
 	addrScheme      string
 	apiAddr         string
+	secure          bool
 }

 // New creates a new ApiClient
@@ -60,6 +62,7 @@ func New(clientID, clientSecret, apiAddr string, secure bool) *ApiClient {
 		tokenAuthConfig: tokenAuthConfig,
 		addrScheme:      addrScheme,
 		apiAddr:         apiAddr,
+		secure:          secure,
 	}
 }

@@ -94,8 +97,8 @@ func (c *ApiClient) RegisterAgent(ctx context.Context, agentName string) (*Agent
 	return info, nil
 }

-// AgentTokenSource returns a token source for a particular agent
-func (c *ApiClient) AgentTokenSource(ctx context.Context, agentID string) oauth2.TokenSource {
+// Credentials returns a token source for a particular agent
+func (c *ApiClient) Credentials(ctx context.Context, agentID string) credentials.PerRPCCredentials {
 	authConfig := &clientcredentials.Config{
 		ClientID:     c.clientID,
 		ClientSecret: c.clientSecret,
@@ -103,5 +106,7 @@ func (c *ApiClient) AgentTokenSource(ctx context.Context, agentID string) oauth2
 		AuthStyle:    oauth2.AuthStyleInHeader,
 	}

-	return authConfig.TokenSource(ctx)
+	ts := authConfig.TokenSource(ctx)
+
+	return newTokenPerRPCCreds(ts, c.secure)
 }
diff --git a/agent/pkg/auth/token_auth.go b/agent/pkg/bcloudapi/creds.go
similarity index 64%
rename from agent/pkg/auth/token_auth.go
rename to agent/pkg/bcloudapi/creds.go
index df331be..9f30874 100644
--- a/agent/pkg/auth/token_auth.go
+++ b/agent/pkg/bcloudapi/creds.go
@@ -1,4 +1,4 @@
-package auth
+package bcloudapi

 import (
 	"context"
@@ -7,23 +7,23 @@ import (
 	"google.golang.org/grpc/credentials"
 )

-// TokenPerRPCCreds satisfies the credentials.PerRPCCredentials interface
+// tokenPerRPCCreds satisfies the credentials.PerRPCCredentials interface
 // attaching an up to date token to each RPC request in the GRPC pipeline.
 // When the token is expired a new token is requested by using the client
 // credentials flow.
-type TokenPerRPCCreds struct {
+type tokenPerRPCCreds struct {
 	secure bool
 	ts     oauth2.TokenSource
 }

-func NewTokenPerRPCCreds(ts oauth2.TokenSource, secure bool) credentials.PerRPCCredentials {
-	return &TokenPerRPCCreds{
+func newTokenPerRPCCreds(ts oauth2.TokenSource, secure bool) credentials.PerRPCCredentials {
+	return &tokenPerRPCCreds{
 		ts:     ts,
 		secure: secure,
 	}
 }

-func (t *TokenPerRPCCreds) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
+func (t *tokenPerRPCCreds) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
 	token, err := t.ts.Token()
 	if err != nil {
 		return nil, err
@@ -34,6 +34,6 @@ func (t *TokenPerRPCCreds) GetRequestMetadata(ctx context.Context, uri ...string
 	}, nil
 }

-func (t *TokenPerRPCCreds) RequireTransportSecurity() bool {
+func (t *tokenPerRPCCreds) RequireTransportSecurity() bool {
 	return t.secure
 }
diff --git a/agent/pkg/registrator/registrator.go b/agent/pkg/registrator/registrator.go
index a77c90d..4ed594d 100644
--- a/agent/pkg/registrator/registrator.go
+++ b/agent/pkg/registrator/registrator.go
@@ -8,7 +8,6 @@ import (
 	pkgk8s "github.com/buoyantio/linkerd-buoyant/agent/pkg/k8s"
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	log "github.com/sirupsen/logrus"
-	"golang.org/x/oauth2"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -37,23 +36,23 @@ func New(bcloudApiClient bcloudapi.Client, k8sAPI *k8s.KubernetesAPI) *Registrat
 // EnsureRegistered performes agent registration if necessary. It inspects the
 // config map containing the agent metadata and decides whether this is a new agent
 // that needs to be registered.
-func (ar *Registrator) EnsureRegistered(ctx context.Context) (*bcloudapi.AgentInfo, oauth2.TokenSource, error) {
+func (ar *Registrator) EnsureRegistered(ctx context.Context) (*bcloudapi.AgentInfo, error) {
 	cm, err := ar.k8sAPI.
 		CoreV1().
 		ConfigMaps(pkgk8s.AgentNamespace).
 		Get(ctx, agentMetadataConfigMap, metav1.GetOptions{})
 	if err != nil {
 		if kerrors.IsNotFound(err) {
-			return nil, nil, fmt.Errorf("could not find %s config map", agentMetadataConfigMap)
+			return nil, fmt.Errorf("could not find %s config map", agentMetadataConfigMap)
 		}
-		return nil, nil, err
+		return nil, err
 	}

 	agentID, hasAgentID := cm.Data[pkgk8s.AgentIDKey]
 	agentName, hasAgentName := cm.Data[pkgk8s.AgentNameKey]

 	if !hasAgentName {
-		return nil, nil, fmt.Errorf("%s config map needs to have an %s key", agentMetadataConfigMap, pkgk8s.AgentNameKey)
+		return nil, fmt.Errorf("%s config map needs to have an %s key", agentMetadataConfigMap, pkgk8s.AgentNameKey)
 	}

 	if hasAgentID {
@@ -64,12 +63,12 @@ func (ar *Registrator) EnsureRegistered(ctx context.Context) (*bcloudapi.AgentIn
 			IsNewAgent: false,
 		}

-		return ai, ar.bcloudApiClient.AgentTokenSource(ctx, agentID), nil
+		return ai, nil
 	}

 	info, err := ar.bcloudApiClient.RegisterAgent(ctx, agentName)
 	if err != nil {
-		return nil, nil, fmt.Errorf("agent registration failed: %w", err)
+		return nil, fmt.Errorf("agent registration failed: %w", err)
 	}

 	cm.Data[pkgk8s.AgentIDKey] = info.AgentID
@@ -78,8 +77,8 @@ func (ar *Registrator) EnsureRegistered(ctx context.Context) (*bcloudapi.AgentIn
 		ConfigMaps(pkgk8s.AgentNamespace).
 		Update(ctx, cm, metav1.UpdateOptions{})
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to update %s configm map: %w", agentMetadataConfigMap, err)
+		return nil, fmt.Errorf("failed to update %s configm map: %w", agentMetadataConfigMap, err)
 	}

-	return info, ar.bcloudApiClient.AgentTokenSource(ctx, info.AgentID), nil
+	return info, nil
 }
diff --git a/agent/pkg/registrator/registrator_test.go b/agent/pkg/registrator/registrator_test.go
index adceafd..16d1100 100644
--- a/agent/pkg/registrator/registrator_test.go
+++ b/agent/pkg/registrator/registrator_test.go
@@ -125,7 +125,7 @@ func TestRegistrator(t *testing.T) {

 			apiClient := bcloudapi.New(clientID, clientSecret, apiAddr, false)
 			registrator := New(apiClient, k8sApi)
-			info, _, err := registrator.EnsureRegistered(context.Background())
+			info, err := registrator.EnsureRegistered(context.Background())
 			if err != nil {
 				if tc.expErr == nil {
 					t.Fatalf("Got unexpected error: %s", err)
```

agent/pkg/auth/token_auth.go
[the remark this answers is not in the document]
siggy: yeah, when `secure == true`, i think it makes sense to use `google.golang.org/grpc`'s implementation, then fall back to our own if `!secure`:
```diff
diff --git a/agent/pkg/auth/token_auth.go b/agent/pkg/auth/token_auth.go
index df331be..83ccb72 100644
--- a/agent/pkg/auth/token_auth.go
+++ b/agent/pkg/auth/token_auth.go
@@ -5,6 +5,7 @@ import (

        "golang.org/x/oauth2"
        "google.golang.org/grpc/credentials"
+       "google.golang.org/grpc/credentials/oauth"
 )

 // TokenPerRPCCreds satisfies the credentials.PerRPCCredentials interface
@@ -13,17 +14,21 @@ import (
 // credentials flow.
 type TokenPerRPCCreds struct {
        secure bool
-       ts     oauth2.TokenSource
+       ts     oauth.TokenSource
 }

 func NewTokenPerRPCCreds(ts oauth2.TokenSource, secure bool) credentials.PerRPCCredentials {
        return &TokenPerRPCCreds{
-               ts:     ts,
+               ts:     oauth.TokenSource{TokenSource: ts},
                secure: secure,
        }
 }

 func (t *TokenPerRPCCreds) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
+       if t.secure {
+               return t.ts.GetRequestMetadata(ctx, uri...)
+       }
+
        token, err := t.ts.Token()
        if err != nil {
                return nil, err
```

agent/pkg/bcloudapi/client.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return info, nil
+}
+
+// AgentTokenSource returns a token source for a particular agent
+func (c *ApiClient) AgentTokenSource(ctx context.Context, agentID string) oauth2.TokenSource {
+	authConfig := &clientcredentials.Config{
+		ClientID:     c.clientID,
+		ClientSecret: c.clientSecret,
+		TokenURL:     fmt.Sprintf("%s://%s%s?agent_id=%s", c.addrScheme, c.apiAddr, agentTokenEndpoint, agentID),
```
siggy, line 102: `k8s.AgentIDKey`

agent/pkg/k8s/k8s.go
```diff
@@ -73,6 +73,13 @@ const (
 	Pod         = "Pod"
 	ReplicaSet  = "ReplicaSet"
 	StatefulSet = "StatefulSet"
+
+	// AgentNamespace is the namespace where the Buoyant Cloud agent is installed.
+	AgentNamespace = "buoyant-cloud"
```
siggy, line 78: the cli already relies on `agent/pkg/k8s`, so i think we could remove `cli/pkg/k8s.Namespace` and reference this one instead.

siggy (commented):
great progress. i left a few diff's that are probably conflict-y with each other, lmkwyt.

agent/main.go
zaharidichev: That is indeed much slicker, thanks

agent/cmd/agent/main.go
```diff
@@ -0,0 +1,191 @@
...
+	"google.golang.org/grpc/credentials"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+
+	// Load all the auth plugins for the cloud providers.
+	// This enables connecting to a k8s cluster from outside the cluster, during
+	// development.
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
```
siggy, line 31: move this into `agent/main.go`

https://github.com/golang/go/wiki/CodeReviewComments#import-blank

siggy (approved):
updates look good! 👍 🚢 

(i'm going to continue hitting "approve" if there are no deal-breaking issues in the latest set of changes, but i'm guessing we'll just keep this PR long-running, and i'll continue to review and approve as we go)

agent/main.go
```diff
@@ -1,178 +1,26 @@
...
-	log.SetLevel(level)
-	log.SetFormatter(&log.TextFormatter{FullTimestamp: true})
-
-	// klog flags for debugging
-	if level >= log.DebugLevel {
-		flag.Set("stderrthreshold", "INFO")
-		flag.Set("v", "6") // At 7 and higher, authorization tokens get logged.
+	if len(os.Args) < 2 {
+		fmt.Println("expected a subcommand")
+		os.Exit(1)
```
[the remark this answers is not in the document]
zaharidichev: Interesting.. what would be the point of that? I guess if we want to do something like that, we can exclude the config map from the agent manifest altogether? But then, what would a fully hydrated manifest look like? I guess we can provide both agentName/agentID as params and based on whether they are set or not, we either perform registration or just create/update the config map. We can try this approach out. WDYT ?

zaharidichev:
@siggy I have updated the docs to explain how to run both the registrator and agent in dev. I think it works fine with piping the config map to kubectl

agent/pkg/flags/flags.go
```diff
@@ -0,0 +1,69 @@
+package flags
+
+import (
+	"flag"
+	"os"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/klog"
```
siggy, line 8: `k8s.io/klog/v2` ?

siggy (approved):
end-to-end works ! 👍 

Screen Shot 2022-02-14 at 11 34 47 AM

agent/cmd/registrator/main.go
```diff
@@ -25,6 +25,7 @@ func Main(args []string) {
 	apiAddr := cmd.String("api-addr", "api.buoyant.cloud:443", "address of the Buoyant Cloud API")
 	kubeConfigPath := cmd.String("kubeconfig", "", "path to kube config")
 	insecure := cmd.Bool("insecure", false, "disable TLS in development mode")
+	agentMetadataMap := cmd.String("agent-metadata-map", registrator.DefaultAgentMetadataConfigMap, "the name of the agent metadata map")
```
siggy, line 28: i think it's ok to define `DefaultAgentMetadataConfigMap` as a private constant in this main, unless we're going to need it in the `registrator` package or elsewhere.

siggy (approved):
`agent-metadata-map` config change looks good, left one comment. 👍 🚢

zaharidichev:
need to fix tests

siggy (approved):
latest updates look good modulo tests 👍 🚢

.github/workflows/actions.yml
```diff
@@ -176,6 +176,11 @@ jobs:
       run: |
         kubectl create ns buoyant-cloud
 
+    - name: Build and load agent image
+      run: |
+        docker build  -f agent/Dockerfile . -t bcloud-agent:ci-run
+        kind load docker-image bcloud-agent:ci-run --name=chart-testing
```
siggy, line 182: can you add a `Cache Docker layers` step, similar to the `Docker Build and Push` job? i think all we need is that one step, i don't think we need the `Move cache` step, since this doesn't use `buildkit`.

### Before
Screen Shot 2022-02-18 at 3 17 45 AM

### After
Screen Shot 2022-02-18 at 3 14 25 AM

siggy (commented):
glad to see the docker build / helm test change was pretty straightforward. left one comment around caching, to speed things up.

related, how about moving the `agentImage` and docker build / helm test changes into a separate PR, and then do another agent release? i'd like to get as many of these changes as possible pushed through the full release process prior to merging this long-running PR.

zaharidichev:
> related, how about moving the agentImage and docker build / helm test changes into a separate PR, and then do another agent release? i'd like to get as many of these changes as possible pushed through the full release process prior to merging this long-running PR.

Makes sense, reverting, #49 should solve the problem

agent/cmd/agent/main.go
```diff
@@ -92,18 +93,20 @@ func Main(args []string) {
...
 	bcloudApiClient := bcloudapi.New(clientID, clientSecret, *apiAddr, secure)
 	perRPCCreds := bcloudApiClient.Credentials(context.Background(), *agentID)
-	opts := []grpc.DialOption{grpc.WithPerRPCCredentials(perRPCCreds)}
-	if *insecure {
-		opts = append(opts, grpc.WithInsecure())
-	} else {
-		creds := credentials.NewTLS(&tls.Config{})
-		opts = append(opts, grpc.WithTransportCredentials(creds))
+
+	tlsCreds := insecure.NewCredentials()
```
siggy, line 100: 👍

siggy (approved):
nice!

siggy (approved):
confirmed `no-tls` work 👍

charts/linkerd-buoyant/templates/bcloud-agent.yaml
```diff
@@ -77,3 +85,25 @@ spec:
...
+          valueFrom:
+            secretKeyRef:
+              key: client_id
+              name: buoyant-cloud-org-credentials
+        - name: CLIENT_SECRET
+          valueFrom:
+            secretKeyRef:
+              key: client_secret
+              name: buoyant-cloud-org-credentials
+
```
siggy, line 109: extra newline

cli/cmd/install.go
```diff
@@ -33,156 +30,99 @@ This command provides the Kubernetes configs necessary to install the Buoyant
...
+			}
+
+			if clientID == "" {
+				return errors.New("install command requires setting CLIENT_ID")
+			}
+			if clientSecret == "" {
+				return errors.New("install command requires setting CLIENT_SECRET")
+			}
+
+			apiClient := bcloudapi.New(clientID, clientSecret, installCfg.bcloudAPI, !installCfg.noTLS)
```
klingerf, line 79: I think this needs to be `installCfg.noTLS` instead of `!installCfg.noTLS`.

I tried running the agent install command from this branch against an AKS cluster, which should require TLS, and it failed with this error:

```
$ ./bin/go-run cli install --context cluster1     
Error: failed to retrieve agent manifest from bcloud server for agent identifier bcloudapi.AgentID XXXXXXXXXXX: Get "http://api.buoyant.cloud:443/agent.yaml?agent_id=XXXXXXXXXXX": oauth2: cannot fetch token: 400 Bad Request
Response: <html>
<head><title>400 The plain HTTP request was sent to HTTPS port</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<center>The plain HTTP request was sent to HTTPS port</center>
<hr><center>nginx</center>
</body>
</html>
```

> The plain HTTP request was sent to HTTPS port

`./bin/go-run cli install --context cluster1 --no-tls` fixes it, but the `--no-tls` flag shouldn't be there.

cli/cmd/install.go
```diff
@@ -33,156 +30,99 @@ This command provides the Kubernetes configs necessary to install the Buoyant
...
+
+			return install(cmd.Context(), installCfg, client, apiClient)
 		},
 	}
 
+	cmd.Flags().StringVar(&installCfg.agentName, "agent-name", "", "The name of the agent.")
```
klingerf, line 85: In the bcloud UI, we refer to these strings as clusters, not agents, so it might be worth renaming this flag to `--cluster-name` (let's discuss at standup).

klingerf (commented):
@zaharidichev I left a few additional comments below aimed at making the waterslide flow a bit easier to navigate. See my other PR for a bit more context.

cli/cmd/install.go
zaharidichev: We can do that, but then do we follow this convention in the code as well? To me cluster-name is a bit confusing as it might be mistaken for the name of the context... We can go either way

klingerf (approved):
⭐ Looks great! I tested a bunch of a new functionality as part of reviewing your other PR, and had just a few minor comments below.

cli/cmd/install.go
```diff
@@ -2,213 +2,130 @@ package cmd
...
+			}
+
+			if clientID == "" || clientSecret == "" {
+				fmt.Fprint(
+					cfg.stderr,
+					"This command requires that the BUOYANT_CLOUD_CLIENT_ID and BUOYANT_CLOUD_CLIENT_SECRET\n",
+					"environment variables be set. To retrieve the correct values for these\n",
+					"variables, visit: https://buoyant.cloud/settings?cli=1.\n",
+				)
+				return fmt.Errorf("no credentials set")
```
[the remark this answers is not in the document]
zaharidichev: You are right, I was not sure which approach I should take as we are not very consistent about it

PR #49, opened 2022-02-18
Build image from source for helm tests

zaharidichev:
This PR changes our GH actions logic to use agent image built from the current source rather than pull the `latest` version.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -194,6 +194,7 @@ jobs:
         file: agent/Dockerfile
         platforms: linux/amd64
         push: false
+        load: true
         tags: ghcr.io/buoyantio/linkerd-buoyant:ci-run
         cache-from: type=local,src=/tmp/.buildx-cache
         cache-to: type=local,dest=/tmp/.buildx-cache-new
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -176,6 +176,19 @@ jobs:
       run: |
         kubectl create ns buoyant-cloud
 
+    - name: Cache Docker layers
+      uses: actions/cache@v2
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/actions.yml
```diff
@@ -176,6 +176,19 @@ jobs:
       run: |
         kubectl create ns buoyant-cloud
 
+    - name: Cache Docker layers
+      uses: actions/cache@v2
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-buildx-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-buildx-
```
siggy, line 185: i don't think i see the cache getting used, the build is taking 3:58:
https://github.com/BuoyantIO/linkerd-buoyant/runs/5251084572?check_suite_focus=true#step:13:12

compare to that to this `main` merge using `build-push-action`, takes 5 seconds:
https://github.com/BuoyantIO/linkerd-buoyant/runs/4713356902?check_suite_focus=true#step:9:143

i think the issue is that caching `/tmp/.buildx-cache` only works if we're using buildkit and specifying the cache path. recommend using `build-push-action` and `Move cache`, set `push: false`, and then do the `kind load` ourselves.

zaharidichev:
> i think the issue is that caching /tmp/.buildx-cache only works if we're using buildkit and specifying the cache path. recommend using build-push-action and Move cache, set push: false, and then do the kind load ourselves.

Switched to buildx. Now it takes 10s. Interesting, you need to explicitly specify that you want the built artifact to be exported to docker before you can use it, otherwise it only stays in cache (or gets pushed if that is enabled).

siggy (approved):
nice! 👍 🚢 
Screen Shot 2022-02-20 at 7 23 25 PM

PR #51, opened 2022-02-21
Move the CLI to the new auth flow

zaharidichev:
This PR changes the CLI to work by using the new auth/autoregistration flow. This means that when agent registration needs to happen, the CLI will not bring you to the Bcloud UI to create the agent but will instead render a manifest that will take care of registration. 

Additionally, the `check` command has been tweaked to check for the presence of the resources that are not required.

### DIFF
--- a/agent/pkg/bcloudapi/client.go
+++ b/agent/pkg/bcloudapi/client.go
@@ -18,7 +18,27 @@ const (
 	agentTokenEndpoint    = "/agent-token"
 	tokenEndpoint         = "/token"
 	registerAgentEndpoint = "/register-agent"
+	agentManifestEndpoint = "/agent.yaml"
 )
+
+// AgentIdentifier identifies an agent when calling the manifest rendering API
+type AgentIdentifier interface {
+	Value() string
+}
+
+// AgentID identifies and agent with its ID
+type AgentID string
+
+func (id AgentID) Value() string {
+	return string(id)
+}
+
+// AgentName identifies and agent with its name
+type AgentName string
+
+func (name AgentName) Value() string {
+	return string(name)
+}
 
 // AgentInfo contains all data to describe an agent that has been
 // registered in Bcloud.
@@ -31,6 +51,7 @@ type AgentInfo struct {
 type Client interface {
 	RegisterAgent(ctx context.Context, agentName string) (*AgentInfo, error)
 	Credentials(ctx context.Context, agentID string) credentials.PerRPCCredentials
+	GetAgentManifest(ctx context.Context, identifier AgentIdentifier) (string, error)
 }
 
 type client struct {
@@ -102,6 +123,47 @@ func (c *client) RegisterAgent(ctx context.Context, agentName string) (*AgentInf
 	}
 
 	return info, nil
+}
+
+// GetAgentManifest renders a manifest for an agent
+func (c *client) GetAgentManifest(ctx context.Context, identifier AgentIdentifier) (string, error) {
+	client := c.tokenAuthConfig.Client(ctx)
+
+	manifestURL := c.base
+	manifestURL.Path = agentManifestEndpoint
+
+	var queryKey string
+	switch identifier.(type) {
+	case AgentID:
+		queryKey = k8s.AgentIDKey
+	case AgentName:
+		queryKey = k8s.AgentNameKey
+	default:
+		return "", fmt.Errorf("invalid agent identifier %+v", identifier)
+	}
+
+	manifestURL.RawQuery = url.Values{queryKey: []string{identifier.Value()}}.Encode()
+
+	req, err := http.NewRequest(http.MethodGet, manifestURL.String(), nil)
+	if err != nil {
+		return "", err
+	}
+	rsp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer rsp.Body.Close()
+
+	if rsp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("agent manifest rendering api returned: %d", rsp.StatusCode)
+	}
+
+	data, err := io.ReadAll(rsp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	return string(data), nil
 }
 
 // Credentials returns a token source for a particular agent
--- a/agent/pkg/bcloudapi/client.go
+++ b/agent/pkg/bcloudapi/client.go
@@ -26,14 +26,14 @@ type AgentIdentifier interface {
 	Value() string
 }
 
-// AgentID identifies and agent with its ID
+// AgentID identifies an agent with its ID.
 type AgentID string
 
 func (id AgentID) Value() string {
 	return string(id)
 }
 
-// AgentName identifies and agent with its name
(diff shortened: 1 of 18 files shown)


agent/pkg/bcloudapi/client.go
```diff
@@ -18,8 +18,28 @@ const (
...
 	registerAgentEndpoint = "/register-agent"
+	agentManifestEndpoint = "/agent.yaml"
 )
 
+// AgentIdentifier identifies an agent when calling the manifest rendering API
+type AgentIdentifier interface {
+	Value() string
+}
+
+// AgentID identifies and agent with its ID
```
siggy, line 29:
```
// AgentID identifies an agent with its ID.
```

agent/pkg/bcloudapi/client.go
```diff
@@ -18,8 +18,28 @@ const (
...
+type AgentID string
+
+func (id AgentID) Value() string {
+	return string(id)
+}
+
+// AgentName identifies and agent with its name
```
siggy, line 36:
```
// AgentName identifies an agent with its name.
```

cli/cmd/uninstall.go
```diff
@@ -62,7 +62,7 @@ func uninstall(ctx context.Context, cfg *config, client k8s.Client) error {
 
 	// if agent is present, output its name and URL for posterity
 	if agent != nil {
-		fmt.Fprintf(cfg.stderr, "Agent manifest will remain available at:\n%s\n\n", agent.URL)
+		fmt.Fprintf(cfg.stderr, "Agent manifest will remain available via :\nlinkerd buoyant install --agent-name=%s\n\n", agent.Name)
```
siggy, line 65: modify comment, a couple lines above here:
```
// if agent is present, output its name and command for posterity
```

cli/pkg/healthcheck/healthcheck.go
```diff
@@ -130,13 +130,24 @@ func (hc *HealthChecker) globalChecks() []healthcheck.Checker {
...
+			Fatal().
+			WithCheck(func(ctx context.Context) error {
+				cm, err := hc.k8s.ConfigMap(ctx)
+				if err != nil {
+					return err
+				}
+
+				return checkLabel(cm.GetLabels(), k8s.PartOfKey, k8s.PartOfVal)
+			}),
+		*healthcheck.NewChecker("buoyant-cloud-org-credentials Secret exists").
```
siggy, line 143: while we're modify check, can you move these two checks to immediately after `buoyant-cloud Namespace has correct labels`? it's surprising to see these checks in between `buoyant-cloud-agent` checks.
```
√ linkerd-buoyant can determine the latest version
√ linkerd-buoyant cli is up-to-date
√ buoyant-cloud Namespace exists
√ buoyant-cloud Namespace has correct labels
√ buoyant-cloud-agent ClusterRole exists
√ buoyant-cloud-agent ClusterRoleBinding exists
√ buoyant-cloud-agent ServiceAccount exists
√ agent-metadata ConfigMap exists
√ buoyant-cloud-org-credentials Secret exists
√ buoyant-cloud-agent Deployment exists
...
```

cli/pkg/k8s/agent.go
```diff
@@ -15,31 +15,29 @@ const (
...
 )
 
 // Agent represents the Buoyant Cloud agent. Any of these fields may not be
 // present, depending on which resources are already on the cluster.
 type Agent struct {
 	Name    string
+	Id      string
 	Version string
-	URL     string
 }
```
siggy, line 28: this feels similar to `AgentIdentifier`, but i suppose `AgentIdentifier` is intended for input/query, and `Agent` is returned by bcloud.

siggy (commented):
works! left a few comments.

cli/cmd/install.go
```diff
@@ -33,156 +30,83 @@ This command provides the Kubernetes configs necessary to install the Buoyant
...
 
-			return install(cmd.Context(), cfg, client, browser.OpenURL)
+			clientID := os.Getenv("CLIENT_ID")
+			if clientID == "" {
+				return errors.New("install command requires setting CLIENT_ID")
+			}
+			clientSecret := os.Getenv("CLIENT_SECRET")
+			if clientSecret == "" {
+				return errors.New("install command requires setting CLIENT_SECRET")
+			}
```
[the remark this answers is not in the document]
zaharidichev: Very nice!

cli/pkg/k8s/agent.go
zaharidichev: The difference is that `AgentIdentifier` is always a oneOf, while with `Agent` we can have both the fields set.  I think we can still reuse it but do not see much of a point.

cli/cmd/install.go
```diff
@@ -33,156 +30,99 @@ This command provides the Kubernetes configs necessary to install the Buoyant
...
+			)
+			return fmt.Errorf("could not find valid agent installation")
+		}
 	}
 
-	body, err := ioutil.ReadAll(resp.Body)
+	manifest, err := apiClient.GetAgentManifest(ctx, identifier)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to retrieve agent manifest from bcloud server for agent identifier %T %s", identifier, identifier.Value())
```
siggy, line 119: let's add the `err` to the output:
```
return fmt.Errorf("failed to retrieve agent manifest from bcloud server for agent identifier %T %s: %w", identifier, identifier.Value(), err)
```

cli/cmd/root.go
```diff
@@ -49,7 +49,10 @@ upgrade, and delete functionality.`,
 
 	// hidden flags
 	root.PersistentFlags().StringVar(&cfg.bcloudServer, "bcloud-server", "https://buoyant.cloud", "Buoyant Cloud server to retrieve manifests from (for testing)")
+	root.PersistentFlags().StringVar(&cfg.bcloudAPI, "bcloud-api", "https://api.buoyant.cloud", "Buoyant Cloud API to retrieve manifests from (for testing)")
```
siggy, line 52: the default needs to be `api.buoyant.cloud:443`. `bcloud-server` is used by `check` and `dashboard` to connect to web URLs, `bcloud-api` is used by install as a `Host` field in a `url.URL`.

cli/cmd/uninstall.go
```diff
@@ -60,9 +60,9 @@ func uninstall(ctx context.Context, cfg *config, client k8s.Client) error {
 		fmt.Fprintf(cfg.stderr, "\n")
 	}
 
-	// if agent is present, output its name and URL for posterity
+	// if agent is present, output its name and command for posterity
 	if agent != nil {
-		fmt.Fprintf(cfg.stderr, "Agent manifest will remain available at:\n%s\n\n", agent.URL)
+		fmt.Fprintf(cfg.stderr, "Agent manifest will remain available via :\nlinkerd buoyant install --agent-name=%s\n\n", agent.Name)
```
siggy, line 65: `via:`

siggy (approved):
lgtm. if all my comments make sense, good to merge into #46. 👍 🚢

PR #52, opened 2022-02-25
Increase metrics scrape interval from 10s to 1m

siggy:


charts/linkerd-buoyant/templates/metrics-agent.yaml
```diff
@@ -18,7 +18,6 @@ data:
     metrics:
       wal_directory: /tmp/wal
       global:
-        scrape_interval: 10s
```
zaharidichev, line 21: Should we set that explicitly to avoid a situation where the default changes from 1m to something else?

zaharidichev (approved):
LGTM

siggy:
Closing in favor of #54.

PR #53, opened 2022-02-28
Try to parse api errors

zaharidichev:
This PR adds a change that enables the bcloud API client to try and parse the error return from the API
in Json format.


siggy (approved):
nice! 👍 🚢

